PR #3, opened 2025-01-13
exposes BackgroundFlushPeriod to public state.Parameters

kjezek:
This PR exposes `BackgroundFlushPeriod` to public `state.Properties`. It allows for configuring frequency of flushing cached data to disk from clients. 

Experiment was conducted with two flush periods configured:
* short - nodes flushed every millisecond 
* long - nodes flushed every second

A stress test was added into the source code.

Results:

* **Short** period:
```
 go test . -cpuprofile=cpu.prof  -v  -tags stress_test -run TestStress_Long_vs_Short_Flush_Period/short
```

image

* **Long** period:
```
 go test . -cpuprofile=cpu.prof  -v  -tags stress_test -run TestStress_Long_vs_Short_Flush_Period/long
```
 
image

### DIFF
--- a/go/state/stress_test.go
+++ b/go/state/stress_test.go
@@ -15,6 +15,7 @@ package state_test
 
 import (
 	"testing"
+	"time"
 
 	"github.com/0xsoniclabs/carmen/go/common"
 	"github.com/0xsoniclabs/carmen/go/state"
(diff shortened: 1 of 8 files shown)


HerbertJordan (approved):
Looks good. 

Did you do some manual end-to-end testing for this change to verify that the property has the desired effect?

kjezek:
added measurements

go/state/stress_test.go
```diff
@@ -131,3 +132,53 @@ func TestStress_CanHandleDeleteOfLargeAccount(t *testing.T) {
...
+		period time.Duration
+	}{
+		"short": {time.Millisecond},
+		"long":  {time.Second},
+	}
+
+	for name, config := range tests {
+		t.Run(name, func(t *testing.T) {
+
+			parameters := state.Parameters{
```
HerbertJordan, line 149: nit: should we limit the cache size to keep the memory footprint low?

go/state/stress_test.go
```diff
@@ -131,3 +132,53 @@ func TestStress_CanHandleDeleteOfLargeAccount(t *testing.T) {
 		})
 	}
 }
+
+func TestStress_Long_vs_Short_Flush_Period(t *testing.T) {
```
HerbertJordan, line 136: This test seems to be intended to collect performance data. 

I think it would be better if this would be a benchmark instead - only called on demand and not part of the standard unit test collection.

Wdyt?

HerbertJordan (dismissed):

go/state/stress_test.go
kjezek: ok, I moved the code and adapted to become a benchmark

HerbertJordan (approved):

go/state/stress_test.go
kjezek: sorry, I missed this comment. Yes, I limited the cache size

HerbertJordan (approved):

PR #5, opened 2025-01-30
tests that code is updatable from public facade

kjezek:
This PR tests that the smart contract code is updatable from the public interface.

### DIFF
--- a/go/carmen/database_test.go
+++ b/go/carmen/database_test.go
@@ -2276,3 +2276,78 @@ func TestDatabase_Export(t *testing.T) {
 		t.Fatalf("cannot close db: %v", err)
 	}
 }
+
+func TestHeadBlockContext_Can_Update_Code(t *testing.T) {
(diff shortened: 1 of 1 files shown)


go/carmen/database_test.go
```diff
@@ -2276,3 +2276,78 @@ func TestDatabase_Export(t *testing.T) {
...
+				}
+			}
+
+			return nil
+		}); err != nil {
+			t.Errorf("failed to add block: %v", err)
+		}
+
+		// the last code must be available at the end of the block
+		if err := db.QueryHeadState(func(context QueryContext) {
```
HerbertJordan, line 2347: Can you also check here or in a follow-up whether the history of code changes is correctly tracked on a per-block level in the archive?

HerbertJordan (approved):
Thanks for adding this, looks good.

cabrador (approved):

go/carmen/database_test.go
kjezek: ok, will add

PR #6, opened 2025-01-31
tests that codes are versioned in blocks

kjezek:
This PR extends tests of the public facade, it checks that codes are correctly versioned in blocks when the codes are updated.


cabrador (approved):

HerbertJordan (approved):
Thanks for adding this!

PR #18, opened 2025-05-02
Add linters

wsodsong:
This PR adds linters to go source code. Enabled linters are errcheck, govet, ineffassign, staticcheck and unused.


psalz:
What is the status of this PR @wsodsong @cabrador ? I am not familiar with the subject matter; are these the same lints used in other projects (e.g. the client)? Have they proven useful there?

facuMH:
https://github.com/0xsoniclabs/sonic/blob/main/.golangci.yml

In Sonic client we use the golang ci tool, consider using it, and there you can see a reference config file.

PR #20, opened 2025-05-19
adds schema configuration for benchmark

kjezek:
This PR extends benchmark configuration of Schema settings


go/database/mpt/tool/benchmark.go
```diff
@@ -77,6 +78,11 @@ var (
 		Name:  "keep-state",
 		Usage: "disables the deletion of temporary data at the end of the benchmark",
 	}
+	schemaFlag = cli.IntFlag{
+		Name:  "schema",
+		Usage: "database scheme to use",
+		Value: 5,
+	}
```
cabrador, line 85: Can you add all of the options please?

cabrador (changes requested):

go/database/mpt/tool/benchmark.go
kjezek: You mean in the usage description?

cabrador (approved):

cabrador (approved):

PR #22, opened 2025-05-21
Partially implement HasEmptyStorage for S3

cabrador:
This PR partially implements `HasEmptyStorage` method for `GoStateS3` and `CppState`.

S3 schema does not allow for iterating over a dataset. This makes it hard to correctly assess if the storage is really empty. For this reason, this PR makes the method always return `true`. 

This method was implemented for compatibility with Ethereum to prevent creation of accounts under certain conditions. These conditions are not effective in Sonic/Fantom chain, and for this reason, the method always pretends the storage is empty.

### DIFF
--- a/go/state/cppstate/cpp_state.go
+++ b/go/state/cppstate/cpp_state.go
@@ -290,7 +290,8 @@ func (cs *CppState) CreateWitnessProof(address common.Address, keys ...common.Ke
 
 func (cs *CppState) HasEmptyStorage(addr common.Address) (bool, error) {
 	// S3 schema is based on directly indexed files without ability to iterate
-	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time.
+	// over a dataset. For this reason, this method is implemented as purely
+	// returning a constant value all the time.
 	return true, nil
 }
 
--- a/go/state/cppstate/cpp_state.go
+++ b/go/state/cppstate/cpp_state.go
@@ -289,7 +289,9 @@ func (cs *CppState) CreateWitnessProof(address common.Address, keys ...common.Ke
 }
 
 func (cs *CppState) HasEmptyStorage(addr common.Address) (bool, error) {
-	panic("CppState does not support HasEmptyStorage operation")
+	// S3 schema is based on directly indexed files without ability to iterate
+	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time.
+	return true, nil
 }
 
 func (cs *CppState) Export(context.Context, io.Writer) (common.Hash, error) {
--- a/go/state/gostate/go_schema3.go
+++ b/go/state/gostate/go_schema3.go
@@ -253,7 +253,8 @@ func (s *GoSchema3) GetCodeHash(address common.Address) (hash common.Hash, err e
 
 func (s *GoSchema3) HasEmptyStorage(addr common.Address) (bool, error) {
 	// S3 schema is based on directly indexed files without ability to iterate
-	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time.
+	// over a dataset. For this reason, this method is implemented as purely
+	// returning a constant value all the time.
 	return true, nil
 }
 
--- a/go/state/gostate/go_schema3.go
+++ b/go/state/gostate/go_schema3.go
@@ -252,7 +252,9 @@ func (s *GoSchema3) GetCodeHash(address common.Address) (hash common.Hash, err e
 }
 
 func (s *GoSchema3) HasEmptyStorage(addr common.Address) (bool, error) {
-	panic("HasEmptyStorage is not implemented for Scheme3")
+	// S3 schema is based on directly indexed files without ability to iterate
+	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time.
+	return true, nil
 }
 
 func (s *GoSchema3) GetHash() (hash common.Hash, err error) {
--- a/go/state/state_test.go
+++ b/go/state/state_test.go
@@ -1085,8 +1085,8 @@ func TestS3_HasEmptyStorage(t *testing.T) {
 			}
 		})
 	}
-
 }
+
 func execSubProcessTest(t *testing.T, dir string, stateImpl string, execTestName string) {
 	path, err := os.Executable()
 	if err != nil {
@@ -1107,7 +1107,6 @@ func execSubProcessTest(t *testing.T, dir string, stateImpl string, execTestName
 // createState creates a state with the given name and directory
 func createState(t *testing.T, name, dir string) state.State {
 	for _, config := range initStates() {
-		fmt.Println(config.name())
 		if config.name() == name {
 			state, err := config.createState(dir)
 			if err != nil {
--- a/go/state/state_test.go
+++ b/go/state/state_test.go
@@ -964,6 +964,119 @@ func TestStateRead(t *testing.T) {
 	}
 }
 
+func TestS3_HasEmptyStorage(t *testing.T) {
+	dir := t.TempDir()
+	st := createState(t, "go-file_s3_none", dir)
+
+	// check non-existent addr
+	isEmpty, err := st.HasEmptyStorage(address1)
+	if err != nil {
+		t.Fatalf("failed to check state: %v", err)
+	}
+	if !isEmpty {
+		t.Errorf("isEmpty should always be true")
+	}
+
+	// create account with no storage set
+	err = st.Apply(1, common.Update{
+		CreatedAccounts: []common.Address{address1},
+	})
+	if err != nil {
+		t.Fatalf("failed to apply state: %v", err)
+	}
+
+	exists, err := st.Exists(address1)
+	if err != nil {
+		t.Fatalf("failed to check account: %v", err)
+	}
+	if !exists {
+		t.Fatal("account does not exist")
+	}
+
+	// check existing addr with empty storage
+	isEmpty, err = st.HasEmptyStorage(address1)
+	if err != nil {
+		t.Fatalf("failed to check state: %v", err)
+	}
+	if !isEmpty {
+		t.Errorf("isEmpty should always be true")
+	}
+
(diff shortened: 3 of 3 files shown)


kjezek (changes requested):
Thanks for this change, but test should be added. The tests are the right place to document via code, what is intention of this method. It should show that it returns a constant value all the time.

Also please update the comment int the code according to my advice, and please also update description of this PR.

I think it is somewhat misleading to say that "a vast refactoring" would be needed. I think there is a conceptual issue with how S3 is designed. In particular, it does not allow the client to iterate over a dataset. But reading the comment, the reader can get impression that is was just too much work to change the code, but I think it would need a new concept using search trees and similar, which Is another story.

go/state/cppstate/cpp_state.go
```diff
@@ -289,7 +289,9 @@ func (cs *CppState) CreateWitnessProof(address common.Address, keys ...common.Ke
 }
 
 func (cs *CppState) HasEmptyStorage(addr common.Address) (bool, error) {
-	panic("CppState does not support HasEmptyStorage operation")
+	// S3 schema is based on directly indexed files without ability to iterate
+	// over a dataset, and thus practically impossible to implement this method
```
kjezek, line 293:
```suggestion
	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time. 
```

go/state/gostate/go_schema3.go
```diff
@@ -252,7 +252,9 @@ func (s *GoSchema3) GetCodeHash(address common.Address) (hash common.Hash, err e
 }
 
 func (s *GoSchema3) HasEmptyStorage(addr common.Address) (bool, error) {
-	panic("HasEmptyStorage is not implemented for Scheme3")
+	// S3 schema is based on directly indexed files without ability to iterate
+	// over a dataset, and thus practically impossible to implement this method
```
kjezek, line 256:
```suggestion
	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time. 
```

go/state/state_test.go
```diff
@@ -964,6 +964,119 @@ func TestStateRead(t *testing.T) {
 	}
 }
 
+func TestS3_HasEmptyStorage(t *testing.T) {
```
kjezek, line 967:
```suggestion
func TestSHasEmptyStorage_Always_Returns_True(t *testing.T) {
```

kjezek: also this test should be stored in `go_schema3_test.go,` because it tests specific feature of this schema

go/state/cppstate/cpp_state.go
```diff
@@ -289,7 +289,9 @@ func (cs *CppState) CreateWitnessProof(address common.Address, keys ...common.Ke
 }
 
 func (cs *CppState) HasEmptyStorage(addr common.Address) (bool, error) {
-	panic("CppState does not support HasEmptyStorage operation")
+	// S3 schema is based on directly indexed files without ability to iterate
+	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time.
```
kjezek, line 293:
```suggestion
	// over a dataset. For this reason, this method is implemented as purely 
	// returning a constant value all the time.
```

go/state/gostate/go_schema3.go
```diff
@@ -252,7 +252,9 @@ func (s *GoSchema3) GetCodeHash(address common.Address) (hash common.Hash, err e
 }
 
 func (s *GoSchema3) HasEmptyStorage(addr common.Address) (bool, error) {
-	panic("HasEmptyStorage is not implemented for Scheme3")
+	// S3 schema is based on directly indexed files without ability to iterate
+	// over a dataset. For this reason, this method is implemented as purely returning a constant value all the time.
```
kjezek, line 256:
```suggestion
	// over a dataset. For this reason, this method is implemented as purely 
	// returning a constant value all the time.
```

go/state/state_test.go
```diff
@@ -984,6 +1107,7 @@ func execSubProcessTest(t *testing.T, dir string, stateImpl string, execTestName
 // createState creates a state with the given name and directory
 func createState(t *testing.T, name, dir string) state.State {
 	for _, config := range initStates() {
+		fmt.Println(config.name())
```
kjezek, line 1110: I guess this was not intended

kjezek (changes requested):
Thanks for these changes, please have a look at a few more comments. 

Also update the description of this PR (as asked before)

go/state/state_test.go
cabrador: not really, it affects both go_state as well as cpp_state

kjezek: ok, I see, the code was different before

go/state/state_test.go
```diff
@@ -964,6 +964,129 @@ func TestStateRead(t *testing.T) {
 	}
 }
 
+func TestS3_HasEmptyStorage(t *testing.T) {
+	for _, config := range initStates() {
+		name := config.name()
```
kjezek, line 969: I think you may use `config.config.Schema == 3` not to rely on strings

go/state/state_test.go
```diff
@@ -964,6 +964,129 @@ func TestStateRead(t *testing.T) {
...
+		if !strings.Contains(name, "s3") {
+			continue
+		}
+		t.Run(name, func(t *testing.T) {
+			dir := t.TempDir()
+			st, err := config.createState(dir)
+			if err != nil {
+				t.Fatalf("Cannot init state: %s, err: %v", name, err)
```
kjezek, line 977:
```suggestion
				t.Fatalf("cannot init state: %s, err: %v", name, err)
```

go/state/state_test.go
```diff
@@ -964,6 +964,129 @@ func TestStateRead(t *testing.T) {
...
+			}
+			// check non-existent addr
+			isEmpty, err := st.HasEmptyStorage(address1)
+			if err != nil {
+				t.Fatalf("failed to check state: %v", err)
+			}
+			if !isEmpty {
+				t.Errorf("isEmpty should always be true")
```
kjezek, line 985:
```suggestion
				t.Errorf("HasEmptyStorage should always return true")
```

go/state/state_test.go
```diff
@@ -964,6 +964,129 @@ func TestStateRead(t *testing.T) {
...
+				t.Fatal("account does not exist")
+			}
+
+			// check existing addr with empty storage
+			isEmpty, err = st.HasEmptyStorage(address1)
+			if err != nil {
+				t.Fatalf("failed to check state: %v", err)
+			}
+			if !isEmpty {
+				t.Errorf("isEmpty should always be true")
```
kjezek, line 1010:
```suggestion
				t.Errorf("HasEmptyStorage should always return true")
```

go/state/state_test.go
```diff
@@ -964,6 +964,129 @@ func TestStateRead(t *testing.T) {
...
+				t.Fatal("account should not exist")
+			}
+
+			// check deleted acc
+			isEmpty, err = st.HasEmptyStorage(address1)
+			if err != nil {
+				t.Fatalf("failed to check state: %v", err)
+			}
+			if !isEmpty {
+				t.Errorf("isEmpty should always be true")
```
kjezek, line 1084:
```suggestion
				t.Errorf("HasEmptyStorage should always return true)
```

go/state/state_test.go
```diff
@@ -964,6 +964,128 @@ func TestStateRead(t *testing.T) {
 	}
 }
 
+func TestHasEmptyStorage_Always_Returns_True(t *testing.T) {
```
kjezek, line 967: Since this test issues several updates to the `State`, it would be much simpler to create an array of updates, and iterate   over this array, in each loop apply the update, and verify that `HasEmptyStorage() == true` all the time.

kjezek (changes requested):
The test is pretty long, but it could be much simpler. It always repeats following sequnce:
1. Apply update
2. Check that `HasEmptyStorage` return true

It would be much more straightforward  to create and array of updates and iterate over this array in a loop. 

Otherwise the change looks good.

go/state/state_test.go
```diff
@@ -964,6 +964,51 @@ func TestStateRead(t *testing.T) {
...
+		if config.config.Schema != 3 {
+			continue
+		}
+		dir := t.TempDir()
+		st, err := config.createState(dir)
+
+		var idx uint64
+		for updateName, update := range updates {
+			testName := fmt.Sprintf("%s_%s", config.name(), updateName)
+			err = st.Apply(idx, update)
```
kjezek, line 993: please inline as:

`if err := ....; err != nil { ... }`

I think it is more idiomatic

kjezek (changes requested):
The sub-test executed within a test should not rely on a shared state. The tests can run in-parallel, or in arbitrary order. Especially, because it iterates a map, the order is not guaranteed. 

In particular,` t.Run(...)`  should be removed, and all should be a single test run

cabrador:
> The sub-test executed within a test should not rely on a shared state. The tests can run in-parallel, or in arbitrary order. Especially, because it iterates a map, the order is not guaranteed.
> 
> In particular,` t.Run(...)` should be removed, and all should be a single test run

I liked running subtests to divide each `State` impl into a single test, so that if it fails, we would exactly know, which failed. Although I removed it.

go/state/state_test.go
```diff
@@ -964,6 +964,48 @@ func TestStateRead(t *testing.T) {
...
+		if config.config.Schema != 3 {
+			continue
+		}
+		dir := t.TempDir()
+		st, err := config.createState(dir)
+		if err != nil {
+			t.Fatalf("unable to create state: %v", err)
+		}
+		var idx uint64
+		for _, update := range updates {
```
kjezek, line 993: you can define index in `for i, update := ...`, and you do not need an extra `idx` variable. It would make the code a tiny bit cleaner

kjezek (approved):
thanks for improving according to the review. Looks good!

PR #23, opened 2025-05-22
makes consistent zero hashes and hash of zeros for witness proof

kjezek:
This PR makes consistent return values for hashes of non-empty and empty accounts in a witness proof.

It is made sure that following values are returned for `storageRoot` 
* a hash of a storage root sub-tree when the account exists, and the storage is non empty
* `Hash(Rlp(EmptyNode))` -- a hash of empty node when the account exists, and its storage is empty
* `0x00...000` -- zero values when account is empty, i.e. the proof could not find account address when creating the proof, the proof is still complete though

For `codeHash` following is assured:
* a hash of byte-code is returned when the account exists, and the code is present
* `Hash([]byte)` -- a hash of empty bytes when the account exist, and its code is not assigned. 
* `0x00..000` -- zero values when account is empty, i.e. the proof could not find account address when creating the proof, the proof is still complete though

### DIFF
--- a/go/carmen/database_test.go
+++ b/go/carmen/database_test.go
@@ -1123,13 +1123,16 @@ func TestDatabase_GetProof_Extract_SubProofs(t *testing.T) {
 	}
 
 	if err := db.AddBlock(0, func(context HeadBlockContext) error {
-		for j := 0; j < numAccounts; j++ {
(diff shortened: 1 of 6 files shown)


cabrador (approved):

jenikd (approved):
Looks good

PR #25, opened 2025-06-19
includes stateDB in memory footprint calculation

kjezek:
this PR updates implementation of memory footprint of the top-level API to include `stateDB` object.

Previously it produced only `state`  (underlaying database layer) footprint as it was ment that the `stateDB` contains only temporal data. It turns out that some objects are long living, and there may be potentially bugs causing some of the elements to grow.

For this reason, it was extended to include this data in the memory footprint. 

Furthermore, `stateDB` was extended to include all its fields in the memory footprint. Although most of the fields do not live only for the duration of one block, a possible bug can cause them to grow, and this needs to be captured as well.

### DIFF
--- a/go/carmen/memory_footprint_test.go
+++ b/go/carmen/memory_footprint_test.go
@@ -16,7 +16,7 @@ import (
 	"testing"
 )
 
-func TestMemoryFootprint(t *testing.T) {
+func TestMemoryFootprint_Contains_Data(t *testing.T) {
 	db, err := openTestDatabase(t)
 	if err != nil {
 		t.Fatalf("cannot open test database: %v", err)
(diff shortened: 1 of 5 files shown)


go/carmen/memory_footprint_test.go
```diff
@@ -36,4 +36,9 @@ func TestMemoryFootprint(t *testing.T) {
 	if !strings.Contains(s, "archive") {
 		t.Error("memory-footprint breakdown does not contain 'archive' keyword even though database contains Archive")
 	}
+
+	if !strings.Contains(s, "liveState") {
```
cabrador, line 40: also add condition that checks that it contains `"database"`?

cabrador (approved):

cabrador (approved):

PR #32, opened 2025-06-28
proof race condition

kjezek:
This PR fixes race condition between updating the archive and creating the witness proof.  

It now correctly uses the `Hash` access right to create the proof, because hashes need to be re-computed.


go/state/gostate/go_schema5_test.go
```diff
@@ -180,3 +182,85 @@ func TestGetNodeCacheConfig_BackgroundFlushPeriod(t *testing.T) {
 		t.Errorf("unexpected background flush period: %v != %v", got, want)
 	}
 }
+
+func TestState_Archive_Query_Proof_While_Updating_Race_Detection(t *testing.T) {
```
HerbertJordan, line 186: This looks like a duplication of the test in the `database_test.go` file. Do we need both?

kjezek: right, removed

PR #36, opened 2025-07-01
Limit live and archive caches on import/export

facuMH:
This PR uses the already declared [mpt/io/live.go](https://github.com/0xsoniclabs/carmen/blob/main/go/database/mpt/io/live.go#L50) import/export cache sizes, when they are called from the archive exported methods. 

This PR is part of https://github.com/0xsoniclabs/sonic-admin/issues/244


kjezek (approved):
looks good thanks

PR #40, opened 2025-07-04
adapts proof verification to use parallel visitors

kjezek:
This PR update proof verification to use parallel node visits. It among other benefits prevents deadlock when the nodes were originally traversed by the View access right, while WitnessProof required Hash access right, used concurrently.


go/database/mpt/io/verification_proof.go
```diff
@@ -262,104 +270,95 @@ func generateUnusedKeys(trie verifiableTrie, number int, address common.Address)
...
 			rootHash:       v.rootHash,
 			trie:           v.trie,
 			currentAddress: n.Address(),
 			storage:        make(map[common.Key]common.Value)}
 
-		if err := v.trie.VisitAccountStorage(n.Address(), &storageVisitor); err != nil || storageVisitor.err != nil {
-			v.err = errors.Join(err, storageVisitor.err)
-			return mpt.VisitResponseAbort
+		// visit the storage of the account
+		if err := visitAll(v.directory, v.config, nodeInfo.Id, &storageVisitor, false); err != nil {
```
HerbertJordan, line 318: If I recall correctly, this will visit all nodes of the trie, passing through all storage slots of all accounts. This means the total runtime of this should be something like `O(#accounts * #total_slots)`, which should be be quite extensive. Is this still running in an acceptable time?

HerbertJordan (commented):
I have a question about the scalability of this approach. Please see the respective comment.

PR #43, opened 2025-07-07
Add explicit permission parameter to visit functions

HerbertJordan:
This PR adds an additional parameter to core `Visit` functions facilitating the specification of the access permissions that should be used while performing the iteration through a trie. The supported options are analog to the modes of the locking [schema](https://github.com/0xsoniclabs/carmen/blob/5b29ffd54847c45ac59be6f5afde3e4c0d6032b9/go/database/mpt/shared/shared.go#L24-L41):
  - read access: grants users to read content fields of the object
  - view access: grants users to read content and hash fields
  - hash access: grants users to read content and write hash fields
  - write access: grants write permission to content and hash fields

This extension facilitates the more accurate specification of node access permissions, enabling more elaborate concurrent access patterns.

### DIFF
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -453,21 +453,26 @@ func (a *ArchiveTrie) Flush() error {
 	return a.addError(err)
 }
 
-func (a *ArchiveTrie) VisitTrie(block uint64, visitor NodeVisitor) error {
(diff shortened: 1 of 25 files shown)


kjezek (dismissed):
thanks

kjezek (approved):

PR #44, opened 2025-07-10
Fix print-out of balances and nonces in updates

HerbertJordan:
This PR fixes a cosmetic issue encountered while debugging some other issues: in Updates, balances got printed as the hex-encoding of their decimal encoding. 

This PR fixes the output format for balances and nonces to use decimal values.


kjezek (approved):
thx

PR #46, opened 2025-07-15
Adapts VerkleTrie from go-ethereum

kjezek:
This PR adapts VerkleTrie implementation from Geth (go-ethereum). It implements the `State` interface, which routes method calls to VerkleTrie as an underlaying data structure. 

It is an experimental implementation, which uses in-memory data storage, and serves for functional comparison to alternative implementations.

### DIFF
--- a/go/database/vt/geth/memory_source.go
+++ b/go/database/vt/geth/memory_source.go
@@ -32,10 +32,7 @@ func NewMemorySource() NodeSource {
 
 func (s *memorySource) Node(owner common.Hash, path []byte, hash common.Hash) ([]byte, error) {
 	key := immutable.NewBytes(path)
-	if node, exists := s.nodes[key]; exists {
-		return node, nil
-	}
-	return nil, nil
+	return s.nodes[key], nil
 }
 
 func (s *memorySource) Set(path []byte, value []byte) error {
--- a/go/database/vt/geth/memory_source.go
+++ b/go/database/vt/geth/memory_source.go
@@ -1,0 +1,53 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package geth
+
+import (
+	"github.com/0xsoniclabs/carmen/go/common/immutable"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// NodeSource is an interface for retrieving and storing nodes in memory.
+// It serves as a base for different node storage implementations, such as
+// in-memory storage, cached storage, or persistent storage.
+// It is itself not intended for real-life usage, as the amount of memory
+// would quickly grow to an unmanageable size.
+type memorySource struct {
+	nodes map[immutable.Bytes][]byte
+}
+
+func NewMemorySource() NodeSource {
+	return &memorySource{
+		nodes: make(map[immutable.Bytes][]byte),
+	}
+}
+
+func (s *memorySource) Node(owner common.Hash, path []byte, hash common.Hash) ([]byte, error) {
+	key := immutable.NewBytes(path)
+	if node, exists := s.nodes[key]; exists {
+		return node, nil
+	}
+	return nil, nil
+}
+
+func (s *memorySource) Set(path []byte, value []byte) error {
+	s.nodes[immutable.NewBytes(path)] = value
+
+	return nil
+}
+
+func (s *memorySource) Flush() error {
+	return nil
+}
+
+func (s *memorySource) Close() error {
+	return s.Flush()
+}
--- a/go/database/vt/geth/memory_source.go
+++ b/go/database/vt/geth/memory_source.go
@@ -21,22 +21,26 @@ import (
 // It is itself not intended for real-life usage, as the amount of memory
 // would quickly grow to an unmanageable size.
 type memorySource struct {
-	nodes map[immutable.Bytes][]byte
+	nodes map[immutable.Bytes]immutable.Bytes
 }
 
-func NewMemorySource() NodeSource {
+func newMemorySource() nodeSource {
 	return &memorySource{
-		nodes: make(map[immutable.Bytes][]byte),
+		nodes: make(map[immutable.Bytes]immutable.Bytes),
 	}
 }
 
 func (s *memorySource) Node(owner common.Hash, path []byte, hash common.Hash) ([]byte, error) {
 	key := immutable.NewBytes(path)
-	return s.nodes[key], nil
+	bytes, exists := s.nodes[key]
+	if !exists {
+		return nil, nil
+	}
+	return bytes.ToBytes(), nil
 }
 
-func (s *memorySource) Set(path []byte, value []byte) error {
-	s.nodes[immutable.NewBytes(path)] = value
+func (s *memorySource) set(path []byte, value []byte) error {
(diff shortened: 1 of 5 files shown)


go/database/vt/geth/memory_source.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// NodeSource is an interface for retrieving and storing nodes in memory.
+// It serves as a base for different node storage implementations, such as
+// in-memory storage, cached storage, or persistent storage.
+// It is itself not intended for real-life usage, as the amount of memory
+// would quickly grow to an unmanageable size.
+type memorySource struct {
+	nodes map[immutable.Bytes][]byte
```
HerbertJordan, line 24: The value could also be an `immutable.Bytes` to
 - prevent accidental modifications
 - enforce ownership on the data through the type system

go/database/vt/geth/memory_source.go
```diff
@@ -0,0 +1,53 @@
...
+		nodes: make(map[immutable.Bytes][]byte),
+	}
+}
+
+func (s *memorySource) Node(owner common.Hash, path []byte, hash common.Hash) ([]byte, error) {
+	key := immutable.NewBytes(path)
+	if node, exists := s.nodes[key]; exists {
+		return node, nil
+	}
+	return nil, nil
```
HerbertJordan, line 38: nit: should be the same as
```suggestion
	return s.nodes[key], nil
```

go/database/vt/geth/state.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// verkleState implements the state.State interface for a verkle trie.
+// It adapts to the VerkleTrie implementation from the Ethereum Geth library.
+// This is a reference implementation to compare with the original Geth.
+type verkleState struct {
+	pointCache *utils.PointCache
+	verkle     *trie.VerkleTrie
+	source     singletonNodeReader
+	codes      map[common.Address][]byte // current Verkle Trie does not support code storage, so we use a map to store codes
```
HerbertJordan, line 66:
```suggestion
	codes      map[common.Address][]byte // current Verkle Trie does not support code retrieval, so we use a map to store codes
```

go/database/vt/geth/state.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func (s *verkleState) SetNonce(address common.Address, nonce common.Nonce) error {
+	account, err := s.getAccount(address)
+	if err != nil {
+		return err
+	}
+
+	account.Nonce = nonce.ToUint64()
+	return s.verkle.UpdateAccount(ethcommon.Address(address), account, 0)
```
HerbertJordan, line 80: Is the last parameter of this function causing the code length to be set to 0 whenever the nonce is updated?

go/database/vt/geth/state.go
```diff
@@ -0,0 +1,297 @@
...
+func (s *verkleState) SetBalance(address common.Address, balance amount.Amount) error {
+	account, err := s.getAccount(address)
+	if err != nil {
+		return err
+	}
+
+	val := balance.Uint256()
+	account.Balance = &val
+
+	return s.verkle.UpdateAccount(ethcommon.Address(address), account, 0)
```
HerbertJordan, line 196: Is this implicitly reducing the account's code length to 0?

go/database/vt/geth/state.go
```diff
@@ -0,0 +1,297 @@
...
+		return err
+	}
+
+	// insert code into the trie
+	if err := s.verkle.UpdateContractCode(ethcommon.Address(address), ethcommon.Hash(codeHash), code); err != nil {
+		return err
+	}
+
+	// put in the local map for retrieval
+	s.codes[address] = code
```
HerbertJordan, line 106: nit: you might want to `bytes.Clone` the code here and when retrieving it to prevent accidental modifications.

go/database/vt/geth/state.go
```diff
@@ -0,0 +1,297 @@
...
+	return commonValue, nil
+}
+
+func (s *verkleState) GetCode(address common.Address) ([]byte, error) {
+	// current Verkle Trie does not support retrieval of codes, i.e. we pick them from the map
+	return s.codes[address], nil
+}
+
+func (s *verkleState) GetCodeSize(address common.Address) (int, error) {
+	code := s.codes[address]
```
HerbertJordan, line 145: The code length should be fetched from the verkle trie, as this is the canonical storage location. The code cache here is only a temporary fix, I assume. By fetching it from the trie it would be aligned with the implementation of `GetCodeHash` below.

HerbertJordan (changes requested):

go/database/vt/geth/node_source.go
```diff
@@ -0,0 +1,46 @@
...
+type singletonNodeReader struct {
+	source NodeSource
+}
+
+func (r singletonNodeReader) NodeReader(stateRoot ethcommon.Hash) (database.NodeReader, error) {
+	return r.source, nil
+}
+
+// get is a convenience method to retrieve the underlying NodeSource.
+func (r singletonNodeReader) get() NodeSource {
```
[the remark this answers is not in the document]
kjezek: ok, renamed method to `getSource`

it seems type `singleNodeReader NodeSource` cannot be used when `NodeSource` is an interface

go/database/vt/geth/memory_source.go
kjezek: ok, good point, changed

go/database/vt/geth/state.go
kjezek: good point, added

go/database/vt/geth/state.go
kjezek: there is problem that in current VerkleTrie in geth there is no method to get code.

go/database/vt/geth/state_test.go
```diff
@@ -0,0 +1,385 @@
...
+
+	balance, err := state.GetBalance(common.Address{})
+	if err != nil {
+		t.Fatalf("failed to get balance for empty account: %v", err)
+	}
+
+	if !balance.IsZero() {
+		t.Errorf("expected zero balance for empty account, got %s", balance)
+	}
+}
```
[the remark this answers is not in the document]
kjezek: I suppose code_hash should be` common.Keccak256([]byte{}) `in this case

go/database/vt/geth/state.go
kjezek: oh, good point, fixed

go/database/vt/geth/state.go
kjezek: good point, fixed

go/database/vt/geth/state_test.go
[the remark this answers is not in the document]
kjezek: ok converted using AI :)

go/database/vt/geth/state_test.go
```diff
@@ -0,0 +1,282 @@
...
+		for j := 0; j < numInsertsPerBlock; j++ {
+			addr := common.Address{byte(j), byte(i), byte(i >> 8)}
+			update.CreatedAccounts = append(update.CreatedAccounts, addr)
+			update.Nonces = append(update.Nonces, common.NonceUpdate{Account: addr, Nonce: common.ToNonce(1)})
+			update.Balances = append(update.Balances, common.BalanceUpdate{Account: addr, Balance: amount.New(uint64(i * j))})
+		}
+		require.NoError(t, state.Apply(uint64(i), update), "failed to apply block %d", i)
+
+		hash, err := state.GetHash()
+		require.NoError(t, err, "failed to get block %d", i)
```
HerbertJordan, line 200: The error message here makes no sense.

kjezek: messages here and elsewhere improved

PR #50, opened 2025-07-18
Add configuration for  VSCode

LorenzSchueler:
This is the fourth and last PR in a series of PRs to set up and configure the Rust crate and related tooling for Carmen S6.

This PR configures:
- the recommended extensions in VSCode for working with Rust
- rust analyzer to use nightly `rustfmt`, enforce imports granularity and use the `crate` prefix
- editorconfig

### DIFF
--- a/rust/.editorconfig
+++ b/rust/.editorconfig
@@ -1,0 +1,12 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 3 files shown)


psalz (approved):
So VSCode can pick up nested .vscode folders? Or is the idea that we open the `rust` folder as the workspace root in VSCode?

LorenzSchueler:
Unfortunately, `.vscode` it is not picked up when we open the root dir, only if we open `rust` directly.

PR #51, opened 2025-07-21
Add cryptographic primitives for Carmen's S6 reference in Go

HerbertJordan:
This PR introduces a set of utility types to handle cryptographic commitments using the [Pedersen schema](https://rareskills.io/post/pedersen-commitment).

The main features are:
- a `Value` type for data to commit to
- a `Commit` function committing to a vector of 256 `Value`s, resulting in a `Commitment`
- an `Update` function to update a single value in the vector a commitment is made for
- an `Open` function to produce a proof for a single value of the vector a `Commitment` is committed to
- a `Verify` function, to verify a proof for a single value against a `Commitment`

### For Reviewers
It is best to review this PR in this order:
- start by reading the `vt/memory/doc.go` to understand the objective of this package; the documentation covers material that is prepared as a follow-up, but the objectives are relevant for this PR
- review `vt/commit/value.go`, since it is the smallest building block
- review `vt/commit/commit.go`, which is the main commitment code
- review `vt/commit/proof.go` for openings and verification tools


go/database/vt/memory/commit/proof_test.go
```diff
@@ -0,0 +1,48 @@
...
+	values := [VectorSize]Value{}
+	for i := range uint64(VectorSize) {
+		values[i] = NewValue(i + 1)
+	}
+
+	commitment := Commit(values)
+	require.True(t, commitment.IsValid())
+
+	for i, value := range values {
+		t.Run(fmt.Sprintf("pos=%d", i), func(t *testing.T) {
```
kjezek, line 30: I think i and value must be copied here when running in parallel
```

i := i
value := value

```

go/database/vt/memory/commit/value_test.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func TestValue_SetBit128(t *testing.T) {
+	require := require.New(t)
+
+	// - starting from zero
+	v := NewValue(0)
+	require.Equal([32]byte{}, v.scalar.Bytes(), "initial value should be zero")
+	v.SetBit128()
+	require.Equal([32]byte{15: 1}, v.scalar.Bytes(), "128th bit should be set")
```
kjezek, line 83: cool, I did not know the syntax `{15: 1}`

go/database/vt/memory/doc.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+// Package memory implements a simple reference Verkle trie (S6) for the Carmen
+// project providing
+//   - a proof-of-concept implementation of the Verkle trie in Carmen
+//   - a executable specification for developers
```
kjezek, line 14:
```suggestion
//   - an executable specification for developers
```

kjezek (changes requested):
Thanks, this looks very good, much more readable than geth implementation. 

I have just one major recommendation, I guess the package `commit` should be better placed in `database/vt/common`, because I think it is not specific to memory implementation, right? If we ever want a file implementation, this package would be used as it is.

go/database/vt/memory/commit/commit.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Identity returns the identity commitment, which is the commitment to a vector
+// of zero values. This is the point at infinity on the Banderwagon curve.
+func Identity() Commitment {
+	return Commitment{point: banderwagon.Identity}
+}
+
+// Commit creates a new commitment to a vector of values.
+func Commit(values [VectorSize]Value) Commitment {
```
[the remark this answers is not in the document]
HerbertJordan: I prefer being explicit about array and slice types. Hiding them behind type definitions is adding mental load to the reviewer due to non-local concepts that need to be considered.

go/database/vt/commit/value.go
```diff
@@ -0,0 +1,54 @@
...
+
+package commit
+
+import (
+	"github.com/crate-crypto/go-ipa/banderwagon"
+)
+
+// Value is a numeric value the commitment can be made of. A commit is a
+// commitment to a vector of these values. The value range is approximately
+// 253 bits, which is just under 32 bytes. Thus, it is not possible to
```
psalz, line 19: Is it really "approximately" 253 bits? How can that be?

go/database/vt/commit/commit.go
```diff
@@ -0,0 +1,112 @@
...
+// can be used as a unique identifier for summarizing the root commitment of a
+// Verkle trie.
+func (p Commitment) Compress() [32]byte {
+	return p.point.Bytes()
+}
+
+// Update creates a new commitment that is the same as the original, except
+// that the value at the given position is updated from old to new.
+func (c Commitment) Update(position byte, old, new Value) Commitment {
+	// This is a very inefficient way to update a commitment.
```
psalz, line 88: I thought this this was the more efficient way of doing it, as opposed to recomputing the entire commitment again?

go/database/vt/commit/value.go
```diff
@@ -0,0 +1,54 @@
...
+// value.
+func NewValue(value uint64) Value {
+	var scalar banderwagon.Fr
+	scalar.SetUint64(value)
+	return Value{scalar: scalar}
+}
+
+// NewValuesFromLittleEndianBytes creates a new value from a 32-byte little-endian
+// byte slice. The value is expanded with zeros to 32 bytes if the input is
+// shorter. Inputs longer than 32 bytes are truncated to 32 bytes.
```
psalz, line 38: Is there a use case where values would be truncated? Or is it simply to not have to return an error?

psalz (approved):
With my limited understanding of both the math and Go - looks good!

go/database/vt/commit/value.go
HerbertJordan: The interface, as it is here, offers the possibility to pass arbitrary byte slices and the documentation should state how these are interpreted. Through this, any user of the API should be aware that excessive elements are simply ignored.

Whether there is a use case or not is less a design concern at this point than to keep it simple, flexible and properly specified.

PR #52, opened 2025-07-22
Add in-memory Verkle trie reference in Go

HerbertJordan:
This PR introduces a simple in-memory reference implementation of a Verkle trie.

The Verkle trie offers a simple key/value store interface, with a `Get` and `Set` function. Internally, it maintains a tree of inner and leaf nodes, and facilitates the computation of Merkle-Tree like commitments using the Pedersen commitment scheme.


kjezek (approved):
Thanks, it looks good. Much more straightforward than eth implementation

kjezek (approved):

PR #53, opened 2025-07-22
Support chain state embedding in verkle trie

HerbertJordan:
This PR adds a `state.State` adapter for a Verkle Trie realising an in-memory version of a Carmen S6 state representation following Ethereum's Verkle Trie encoding rules.


go/database/vt/memory/state.go
```diff
@@ -0,0 +1,183 @@
...
+
+// NewState creates a new, empty in-memory state instance.
+func NewState() *State {
+	return &State{
+		trie: &trie.Trie{},
+	}
+}
+
+func (s *State) Exists(address common.Address) (bool, error) {
+	return false, fmt.Errorf("this is not supported by Verkle Tries")
```
kjezek, line 41: We could determine existence from the `basicData`, if it is non-empty, the account exists.

go/database/vt/memory/state.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (s *State) GetBalance(address common.Address) (amount.Amount, error) {
+	key := getBasicDataKey(address)
+	value := s.trie.Get(key)
+	return amount.NewFromBytes(value[16:32]...), nil
```
kjezek, line 47: perhaps it would be useful to have a layer between `State` and `Trie`, which would convert to/from `Account` with fields for `Balance`, `Nonce`, etc.

go/database/vt/memory/state_test.go
```diff
@@ -0,0 +1,559 @@
...
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package memory
```
kjezek, line 11: I guess I should generalise these tests for both `memory` and `geth` impl. Otherwise I would end up writing almost the same tests

kjezek (approved):
Thanks, it looks good. 

I am a bit puzzled that I think Geth hashes Addresses, while `State` implementation here does not, but the tests comparing with Geth are provided. Is it that the commitment is still the same when that data are same, but the structure of the trie differs? 

In any case, should not we hash addresses to have the same implementation as in Geth?

go/database/vt/memory/code.go
```diff
@@ -0,0 +1,59 @@
...
+// of the data section of a push instruction.
+func splitCode(code []byte) []chunk {
+	const PUSH1 = 0x60
+	const PUSH32 = 0x7f
+	isCode := make([]bool, len(code)+33)
+	for i := 0; i < len(isCode); i++ {
+		isCode[i] = true
+		if i < len(code) {
+			cur := code[i]
+			if PUSH1 <= cur && cur <= PUSH32 {
```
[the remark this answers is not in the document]
HerbertJordan: There are 32 different push PUSH OpCodes, which are all in the range of op-code 0x60 to 0x7F (see [here](https://www.evm.codes/)). There are no other ob-codes in this range.

go/database/vt/memory/state.go
HerbertJordan: I would prefer to remove this function from the state interface and let the existence be tested by the transaction context by querying the balance or nonce. I suspect that this function is not actually needed.

go/database/vt/memory/state_test.go
HerbertJordan: The tests are kinda fitted to match the implementation and to check all the code that is covered in this package.

In the case of the `geth` package, which I would assume to be ultimately a thin adapter, an integration test as done at the bottom of this file should be sufficient.

go/database/vt/memory/state.go
```diff
@@ -0,0 +1,183 @@
...
+		trie: &trie.Trie{},
+	}
+}
+
+func (s *State) Exists(address common.Address) (bool, error) {
+	return false, fmt.Errorf("this is not supported by Verkle Tries")
+}
+
+func (s *State) GetBalance(address common.Address) (amount.Amount, error) {
+	key := getBasicDataKey(address)
```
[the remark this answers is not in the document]
HerbertJordan: The function `getTrieKey` performs the hashing of addresses. However, in the Verkle trie addresses are not hashed using Keccak but the Pedersen hash function -- which is essentially computing a commitment point and deriving a hash from this point.

The reason for this is likely that the Pedersen hash function is more efficient in zero-knowledge proofs than Keccak.

go/database/vt/memory/state.go
HerbertJordan: I think this would weaken readability by introducing a level of indirection compared to the current version.

go/database/vt/memory/state.go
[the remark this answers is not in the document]
kjezek: ah, right, I missed that hashing is happening in `getTrieKey`

kjezek (approved):

PR #54, opened 2025-07-24
Release memory footprint buffer by call to foreign language

LorenzSchueler:
This PR adds the function `Carmen_ReleaseMemoryFootprintBuffer` to the FFI interface. This enables us to use different memory allocators.
Before, the buffer was released in Go by a call to `C.free`, however that only works if it was allocated by a call to `malloc` which might not necessarily be the case.


HerbertJordan (approved):
Looks good, thanks!

kjezek (approved):
I rely on @HerbertJordan , I do not very understand this adapter

PR #55, opened 2025-07-24
Duplicate FFI function declarations for each language

LorenzSchueler:
This PR duplicates all function declarations of the FFI interface for each language and adds a corresponding prefix to the function name. This is needed to be able to link both the C++ and Rust implementation without name conflicts.


cpp/state/c_state.h
```diff
@@ -17,6 +17,11 @@
 extern "C" {
 #endif
 
+// Macro to duplicate function declarations for Rust and CPP
+#define DUPLICATE_FOR_LANGS(ret_type, fn) \
```
HerbertJordan, line 21: This macro should be undefined at the end of the file.

cpp/state/c_state.h
```diff
@@ -17,6 +17,11 @@
...
+  ret_type Rust_##fn;                     \
+  ret_type CPP_##fn;
```
HerbertJordan, line 23: nit: "CPP_" => "Cpp_"

HerbertJordan, line 23: Structurally, "Carmen_Rust_XXX" and "Carmen_Cpp_XXX" would be more suiting. You could include the "Carmen_" prefix here and remove it from the functions - no need to try to past that in half-way through the function name.

Wdty?

HerbertJordan (changes requested):
Looks good. The macro should be undefined at the end of the header. Additionally, I added one or two ideas for improving the naming schema. Please have a look, let me know what you think.

HerbertJordan (approved):
Thanks for applying the changes, looks good!

PR #56, opened 2025-07-28
Export FFI functions from Rust

LorenzSchueler:
This is the first PR in a series of PRs to expose all required functions from Rust, and call the safe interface from those functions.

This PR:
- generates bindings from the C header file using [bindgen](https://crates.io/crates/bindgen)
  - Strictly speaking, we don't need any of the function declarations because we are not calling another language from Rust, but rather, Rust from another language. However, they are useful for testing that our exported functions have the correct signatures.
- exports all functions of the FFI interface
- adds compile time checks that the signatures are correct
  - for this check, we pass function pointers for our exported function, and the function declared in the generated bindings to `const fn assert_same_signature<T>(a: T, b: T)`. This way we ensure that both functions have the same signature, because otherwise they would not have the same type `T` which in turn would cause compilation to fail.
  
Note: the C header file is this one `cpp/state/c_state.h`
  
New dependencies:
- [bindgen](https://crates.io/crates/bindgen) the default crate for generating bindings from C header files

Part of https://github.com/0xsoniclabs/sonic-admin/issues/283

### DIFF
--- a/rust/src/ffi/exported.rs
+++ b/rust/src/ffi/exported.rs
@@ -1,0 +1,244 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+#![allow(unused_variables)]
+
+use std::ffi::{c_char, c_int, c_void};
+
+use crate::ffi::bindings;
+
+/// Opens a new state object based on the provided implementation maintaining
+/// its data in the given directory. If the directory does not exist, it is
+/// created. If it is empty, a new, empty state is initialized. If it contains
+/// state information, the information is loaded.
+///
+/// The function returns an opaque pointer to a state object that can be used
+/// with the remaining functions in this file. Ownership is transferred to the
+/// caller, which is required for releasing it eventually using Carmen_Rust_ReleaseState.
+/// If for some reason the creation of the state instance failed, a nullptr is
+/// returned.
+#[unsafe(no_mangle)]
+unsafe extern "C" fn Carmen_Rust_OpenState(
+    schema: u8,
+    state: bindings::StateImpl,
+    archive: bindings::ArchiveImpl,
+    directory: *const c_char,
+    length: c_int,
+) -> *mut c_void {
+    unimplemented!()
+}
+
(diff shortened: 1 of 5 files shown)


LuisPH3 (approved):
:+1

rust/src/ffi/exported.rs
```diff
@@ -0,0 +1,244 @@
...
+        Carmen_Rust_GetMemoryFootprint as unsafe extern "C" fn(_, _, _),
+        bindings::Carmen_Rust_GetMemoryFootprint,
+    );
+    assert_same_signature(
+        Carmen_Rust_ReleaseMemoryFootprintBuffer as unsafe extern "C" fn(_, _),
+        bindings::Carmen_Rust_ReleaseMemoryFootprintBuffer,
+    );
+};
+
+const fn assert_same_signature<T>(a: T, b: T) -> (T, T) {
```
psalz, line 242: I was wondering whether the return type is even needed: It is, but I have no idea why :shrug:

psalz (approved):

rust/src/ffi/exported.rs
LorenzSchueler: The problem is that we take T by value. T might implement `Drop`. But `Drop::drop` is not a `const fn`. Therefore, we can not drop `T` in const contexts. But returning it is fine because then it is again the callers' responsibility to drop T. At the call side we know the specific type of `T` (function pointers) and know that `T` does not implement `Drop`.
Another option would be to take `T` by reference, but this is uglier on the caller side.

rust/src/ffi/exported.rs
```diff
@@ -0,0 +1,244 @@
...
+unsafe extern "C" fn Carmen_Rust_GetArchiveState(state: *mut c_void, block: u64) -> *mut c_void {
+    unimplemented!()
+}
+
+/// Gets the current state of the given account.
+#[unsafe(no_mangle)]
+unsafe extern "C" fn Carmen_Rust_GetAccountState(
+    state: *mut c_void,
+    addr: *mut c_void,
+    out_state: *mut c_void,
```
HerbertJordan, line 70: Rust question: would [type aliases](https://doc.rust-lang.org/reference/items/type-aliases.html) for the various object types work here? It would help readability a bit, but would it be Rusty?

The type for the address may, for instance, be defined as a pointer to a 20-byte array. Balances are pointers to 16-byte arrays, and nonces to an 8-byte array. This would be more precise then the generic void pointer.

I understand that the typing is not enforced on the C interface, but it may help adding some type checks to implementations.

HerbertJordan (approved):

PR #57, opened 2025-07-28
Add type aliases and implement block update parsing

LorenzSchueler:
This is the second PR in a series of PRs to expose all required functions from Rust, and call the safe interface from those functions.

This PR adds type aliases (for `Address`, `Key`, ...) and the `Update` type along with all the types of its fields. This type holds all updates that should be applied to a block.
It also implements parsing `Update` from its encoded form.
Unlike in C++ which copies the data, this PR implements zero-copy deserialization where possible.
  
Relevant Go/ C++ code:
- `cpp/state/update.cc` (Update::FromBytes)
- `go/common/update.go` (ToBytes)
  
New dependencies:
- [zerocopy](https://crates.io/crates/zerocopy) provides traits and macros to do SAFE `transmute`s (they perform compile time analysis to make sure that this is safe), which is very useful for zero cost conversions. This is not only useful here, but will also be useful when converting between byte slices stored on disk and their in memory representation.

Part of https://github.com/0xsoniclabs/sonic-admin/issues/283


rust/src/types/update.rs
```diff
@@ -0,0 +1,508 @@
...
+                },
+                SlotUpdate {
+                    addr: [23; 20],
+                    key: [24; 32],
+                    value: [25; 32],
+                },
+            ],
+        };
+
+        let writes = [
```
ihtefyw, line 417: This test is hard to follow. 
Please add some comments to describe what each write does and the what the loop at the end is doing

ihtefyw (changes requested):
Looks good, just some minor hints on hints on testing

ihtefyw (dismissed):

psalz (dismissed):
Looks solid! I noticed that `Hash`, `AccountState` and the two constants `ACCOUNT_STATE_UNKNOWN` and `ACCOUNT_STATE_EXISTS` are currently unused. Should we delay introducing these until they are needed to make them easier to connect in a blame?

PR #58, opened 2025-07-28
Add safe CarmenS6 trait and call safe interface from unsafe one

LorenzSchueler:
This is the third and last PR in a series of PRs to expose all required functions from Rust, and call the safe interface from those functions.

This PR:
- adds the safe carmen interface (`trait CarmenS6Db`)
- adds `DbState` which will be our main type which implements `CarmenS6Db` (currently all methods are unimplemented, but we need this type already to be able to return a type from `open_carmen_s6_db`
- implements the conversion from the types passed via FFI (mostly `void` pointers) to references of the correct types and calls to the safe interface. This also includes safety comments.
- tests the unsafe interface
  - create is tested by checking that the returned value is not null
  - all other methods are called with a wrapped pointer to `MockCarmenS6Db` on which the corresponding expectation is set
  - these tests are especially useful when run with MIRI, as this checks that the unsafe operations did not cause any UB
- adds a CI step to run [MIRI](https://github.com/rust-lang/miri), which is a tool to detect undefined behavior in Rust.
  
New dependencies:
- [mockall](https://crates.io/crates/mockall) mocking library to derive mock structs from trait definitions

Disclaimer: I know this PR looks huge, but most of it is just documentation, which is also very similar between the functions. IMO, splitting this PR further makes little sense because:
- implementing the unsafe interface needs the safe interface
- the safe interface only makes sense when considering how it gets called
- the unsafe interface should not be merged without running MIRI

Closes https://github.com/0xsoniclabs/sonic-admin/issues/283

### DIFF
--- a/rust/src/ffi/exported.rs
+++ b/rust/src/ffi/exported.rs
@@ -89,8 +89,8 @@ unsafe extern "C" fn Carmen_Rust_OpenState(
 /// - `state` must be valid for reads and writes for the duration of the call
 /// - `state` must not be mutated for the duration of the call
 /// - `state.state` must be a valid pointer to a `dyn CarmenS6Db`
-/// - `state.state` must be valid for reads and writes for the duration of the lifetime of token
-/// - `state.state` must not be mutated for the duration of the lifetime of token
+/// - `state.state` must be valid for reads and writes for the duration of the lifetime of `token`
+/// - `state.state` must not be mutated for the duration of the lifetime of `token`
 #[unsafe(no_mangle)]
 unsafe extern "C" fn Carmen_Rust_Flush(state: *mut c_void) {
     let token = LifetimeToken;
@@ -104,9 +104,9 @@ unsafe extern "C" fn Carmen_Rust_Flush(state: *mut c_void) {
     let state = unsafe { ref_mut_from_ptr_scoped(state as *mut StateWrapper, &token) };
     // SAFETY:
     // - `state.state` is a valid pointer to a `dyn CarmenS6Db` (precondition)
-    // - `state.state` is valid for reads and writes for the duration of the lifetime of token
+    // - `state.state` is valid for reads and writes for the duration of the lifetime of `token`
     //   (precondition)
-    // - `state.state` is not mutated for the duration of the lifetime of token (precondition)
+    // - `state.state` is not mutated for the duration of the lifetime of `token`(precondition)
     let db_state = unsafe { state.to_ref_mut_scoped(&token) };
     #[allow(unused_variables)]
     if let Err(error) = db_state.flush() {
@@ -122,8 +122,8 @@ unsafe extern "C" fn Carmen_Rust_Flush(state: *mut c_void) {
 /// - `state` must be valid for reads and writes for the duration of the call
 /// - `state` must not be mutated for the duration of the call
 /// - `state.state` must be a valid pointer to a `dyn CarmenS6Db`
-/// - `state.state` must be valid for reads and writes for the duration of the lifetime of token
-/// - `state.state` must not be mutated for the duration of the lifetime of token
+/// - `state.state` must be valid for reads and writes for the duration of the lifetime of `token`
+/// - `state.state` must not be mutated for the duration of the lifetime of `token`
 #[unsafe(no_mangle)]
 unsafe extern "C" fn Carmen_Rust_Close(state: *mut c_void) {
     let token = LifetimeToken;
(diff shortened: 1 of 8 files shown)


LuisPH3 (dismissed):
This PR adds the abstraction to invoke the database utilities. and some tooling around it. 

This pr is very large, some type conversions could have been presented in isolation (i.e. integers to enum with bounds check, CI integration should be a separate pr to avoid clutter)

psalz (approved):
Whew!

LuisPH3 (approved):

rust/src/lib.rs
```diff
@@ -7,8 +7,140 @@
...
 mod error;
 mod ffi;
 mod types;
+
+/// Opens a new [CarmenS6Db] database object based on the provided implementation maintaining
+/// its data in the given directory. If the directory does not exist, it is
+/// created. If it is empty, a new, empty state is initialized. If it contains
+/// state information, the information is loaded.
+pub fn open_carmen_s6_db(
+    _schema: u8,
```
HerbertJordan, line 24: The `_schema` parameter actually determines what "sX" version should be produced. So this function is intended to be used for all kind of different schemas. I would thus suggest to
- rename the function to `open_carmen_db`
- rename the type `CarmenS6Db` trait to `CarmenDb` since the trait contains nothing S6 specific and the FFI is in general schema-independent

If there is no support for a given schema on the Rust side (schemas 1-5, >=7), an error should be returned.

Wdyt?

rust/src/ffi/exported.rs
```diff
@@ -31,142 +49,843 @@ unsafe extern "C" fn Carmen_Rust_OpenState(
...
+        Self {
+            state: Box::into_raw(state),
+        }
+    }
+}
+
+/// A zero sized type, used to enforce the correct lifetime for references created from
+/// pointers obtained via FFI. It is assumed that the lifetime of the pointer is bound by the
+/// lifetime of the borrow of [`LifetimeToken`].
+/// When the token is crated at the beginning of a function call, it ensures that the
```
HerbertJordan, line 816:
```suggestion
/// When the token is created at the beginning of a function call, it ensures that the
```

HerbertJordan (dismissed):
I have a very limited understanding of the Rust code, but interface wise it looks good to me. However, you might want to consider dropping the S6 part of the trait and function names.

rust/src/lib.rs
LorenzSchueler: Sounds good

LorenzSchueler:
@kjezek @HerbertJordan I renamed the trait to `CarmenDb` and the S6 implementation to `CarmenS6Db`. Also, the open function is now called `open_carmen_db` and returns an error if the schema is not `6`, as suggested.
@kjezek regarding opening live and archive state, there will be a follow-up PR that clean up the overall FFI interface (see #62) where the database is decoupled from the live state. With this PR I will then also open both states in open db.

HerbertJordan (approved):

ihtefyw (approved):

rust/src/lib.rs
```diff
@@ -7,8 +7,143 @@
...
+) -> Result<Box<dyn CarmenDb>, Error> {
+    if schema != 6 {
+        return Err(Error::UnsupportedSchema(schema));
+    }
+    // here we would open the specific live and archive implementations of CarmenDb based on the
+    // state and archive impls.
+    Ok(Box::new(CarmenS6Db))
+}
+
+/// The safe Carmen interface.
```
kjezek, line 37: is it clear in rust context what `safe` means?

rust/src/types/mod.rs
```diff
@@ -12,14 +12,14 @@ pub use update::Update;
 
 mod update;
 
-/// The Carmen S6 database state implementation.
+/// The Carmen database state implementation.
 pub enum StateImpl {
```
kjezek, line 16: I am not sure what convention in rust is, but reading `...Impl` I did not notice it is actually an enum, not an implementation of something.

kjezek (approved):
It looks good as far as I can see. I asked some more questions about naming conventions

rust/src/lib.rs
LorenzSchueler: At least in conjunction with the line below that states that this gets called from the FFI interface (which is inherently unsafe) it should be clear.

rust/src/types/mod.rs
LorenzSchueler: The naming comes mainly from the C header which calls it that way. I guess it also makes sense, because using this enum you choose which implementation is choosen.

PR #60, opened 2025-07-29
Configure bazel to emit compile_commands.json during build

LorenzSchueler:
This PR configures Bazel to emit a `compile_commands.json` during its build. This is needed for LSP support.
The [bazel docs](https://bazel.build/install/ide#autocomplete-for-source-code) list two approaches, one where you run a command manually, and one where you configure bazel to emit the file automatically. I opted for the later one because this way everything is set up for you once you build the project.


psalz (approved):
If that is what it takes :shrug:

ihtefyw (approved):

PR #61, opened 2025-07-30
cross-checks verkle state implementations

kjezek:
This PR creates a new set of tests of the state, and a new comparing state.

It captures a set of State implementations and runs the test against this test. It checks if the results obtained from each state are the same.

It is supposed to check various implementations that they produce consistent results.


go/database/vt/state_test.go
```diff
@@ -0,0 +1,486 @@
...
+	state, err := initTestedStates(t)
+	require.NoError(t, err, "failed to initialize tested states")
+	defer func() {
+		require.NoError(t, state.Close(), "failed to close state")
+	}()
+
+	addr := common.Address{1}
+
+	update := common.Update{}
+	update.CreatedAccounts = append(update.CreatedAccounts, addr)
```
HerbertJordan, line 32: I have been running a test removing the `CreateAccounts` field from the `Update` struct, and the Sonic block history could be successfully replayed. Also, when dumping and [restoring](https://github.com/0xsoniclabs/carmen/blob/5434be3ba2df3e74c557c129d503550a2aa6ed13/go/database/mpt/io/archive.go#L272-L363) archive state we use deltas between states and we do not require calls to the creation of an account.

I thus suspect that we do not really need the `CreateAccounts` field in the `Update` struct and we do not have to worry about its implementation. We should actually attempt to remove it.

go/database/vt/state_test.go
```diff
@@ -0,0 +1,486 @@
...
+			expectedValue := common.Value{byte(i), byte(i >> 8), byte(j), byte(j >> 8)}
+
+			value, err := state.GetStorage(addr, key)
+			require.NoError(t, err, "failed to get storage for account %x and key %x", addr, key)
+			require.Equal(t, expectedValue, value, "storage value mismatch for account %x and key %x", addr, key)
+		}
+	}
+}
+
+func TestState_Code_Can_Set_And_Receive(t *testing.T) {
```
HerbertJordan, line 159: The ability to update codes (also with different lengths) would be great to be tested as well. Since codes are embedded in the tries, it would be interesting to see whether old codes are truncated or remain partially present when being replaced with shorter codes.

go/database/vt/state_test.go
```diff
@@ -0,0 +1,486 @@
...
+func (s *comparingState) put(name string, state state.State) {
+	s.states[name] = state
+}
+
+func (s *comparingState) Exists(address common.Address) (bool, error) {
+	s.t.Helper()
+	return getCmpStateValue(s, func(name string, state state.State) (bool, error) {
+		return state.Exists(address)
+	})
+}
```
HerbertJordan, line 276: This function is not used in tests.

go/database/vt/state_test.go
```diff
@@ -0,0 +1,486 @@
...
+		return state.GetHash()
+	})
+}
+
+func (s *comparingState) HasEmptyStorage(addr common.Address) (bool, error) {
+	s.t.Helper()
+	return getCmpStateValue(s, func(name string, state state.State) (bool, error) {
+		return state.HasEmptyStorage(addr)
+	})
+}
```
HerbertJordan, line 327: This function is not used in the tests and I think a Verkle Trie might not be able to answer that.

go/database/vt/state_test.go
```diff
@@ -0,0 +1,486 @@
...
+//
+//	I/O features
+//
+
+func (s *comparingState) Flush() error {
+	s.t.Helper()
+	return s.action(func(name string, state state.State) error {
+		return state.Flush()
+	})
+}
```
HerbertJordan, line 430: These functions are not really needed for the given tests.

HerbertJordan (changes requested):
Thanks for adding those tests covering basic consistency between implementations. I have a few suggestions though:
- the tests contain a bit of dead code that could be removed;
- as discussed in other contexts, I suspect that "CreateAccount" is redundant and could be eliminated from the interface, and thus eliminating the need to test for it, and to reflect its special semantic in the `geth` implementation.

The idea is: I would like to introduce implementation artefacts of `geth` into the requirements of the Verkle Trie implementation. We should keep the specification as lean and clean as possible.

go/database/vt/state_test.go
kjezek: ok, removed

go/database/vt/state_test.go
kjezek: good idea, extended

go/database/vt/geth/state_test.go
HerbertJordan: Was this missed by #286?

go/database/vt/state_test.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	gethSt, err1 := geth.NewState(state.Parameters{})
+	refSt, err2 := reference.NewState(state.Parameters{})
+	innerSt, err3 := reference.NewState(state.Parameters{})
+	flatSt, err4 := flat.NewState(t.TempDir(), innerSt)
+
+	st.put("memory", refSt)
+	st.put("geth", gethSt)
+	st.put("flat", flatSt)
```
HerbertJordan: For maintainability, these states should be derived from the variant registry. Everything that declares to be an S6 implementation should be covered.

go/database/vt/state_test.go
```diff
@@ -0,0 +1,417 @@
...
+// getCmpStateValue is a helper function that iterates over all states in the comparingState
+// and applies the provided function to each state. It collects errors and checks
+// for value consistency across all states. If all states return the same value,
+// it returns that value; otherwise, it returns an error indicating the mismatch.
+func getCmpStateValue[T comparable](s *comparingState, do func(name string, state state.State) (T, error)) (T, error) {
+	s.t.Helper()
+	var errs []error
+	var ref *T
+	var refName string
+	for name, state := range s.states {
```
HerbertJordan: Since iteration through maps is random, this will result in the reference version being a different version in every call of `getCmpStateValue`. This could be an annoying property when debugging failing tests.

Can you make a fixed implementation the reference implementation?

HerbertJordan (approved):
Thanks for the improvements!

PR #64, opened 2025-07-30
Require use of at least one inner node in every non-empty Verkle trie

HerbertJordan:
This PR supports the implicit enforcement of an inner node in every non-empty Verkle trie.

### DIFF
--- a/go/database/vt/memory/trie/trie.go
+++ b/go/database/vt/memory/trie/trie.go
@@ -35,13 +35,6 @@ type Trie struct {
 	root node
 }
 
-// NewTrie creates a new empty Verkle trie.
-// The root node is initialized to an empty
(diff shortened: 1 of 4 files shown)


go/database/vt/memory/trie/trie.go
```diff
@@ -55,7 +48,7 @@ func (t *Trie) Get(key Key) Value {
 // already exists, its value will be updated.
 func (t *Trie) Set(key Key, value Value) {
 	if t.root == nil {
-		t.root = newLeaf(key)
+		t.root = &inner{}
```
kjezek, line 51: right, it is much cleaner

kjezek (approved):
tis is better, thanks for this idea

PR #65, opened 2025-07-31
initialises empty accounts with eth empty hash

kjezek:
this PR updates in-memory Verkle Trie implementation to match geth implementation. 

It stores hash of empty code to each created account.  Furthermore, methods to create accounts were updated to be no-op when an account already exists in both geth and memory implementations. 

This change is rather a complication of code with not clear benefit. It is introduced only to be compatible with Ethereum.

### DIFF
--- a/go/database/vt/memory/state.go
+++ b/go/database/vt/memory/state.go
@@ -11,6 +11,7 @@
 package memory
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"fmt"
@@ -22,6 +23,7 @@ import (
 	"github.com/0xsoniclabs/carmen/go/common/witness"
 	"github.com/0xsoniclabs/carmen/go/database/vt/memory/trie"
 	"github.com/0xsoniclabs/carmen/go/state"
+	"github.com/ethereum/go-ethereum/core/types"
 )
 
 // State is an in-memory implementation of a chain-state tracking account and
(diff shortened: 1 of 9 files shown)


go/database/vt/memory/state.go
```diff
@@ -14,6 +14,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"github.com/ethereum/go-ethereum/core/types"
```
HerbertJordan, line 17: nit: somehow your editor seems to use a different order for the imports. It is not grouping them into the standard-library section and the third-party section. Can you try to figure out why this is the case and whether you can get this somehow fixed?

go/database/vt/memory/state.go
```diff
@@ -85,23 +86,34 @@ func (s *State) HasEmptyStorage(addr common.Address) (bool, error) {
 }
 
 func (s *State) Apply(block uint64, update common.Update) error {
+	
+	maybeEmptyCodeAccounts := make(map[common.Address]struct{})
```
HerbertJordan, line 90: For the big picture, I think the responsibility on what to initialize and what not should be the job of the block processor and/or the transaction context, not the state. If a new account is created, the update should contain a corresponding code update setting the account's code to the empty code.

The update contains a [CreatedAccount](https://github.com/0xsoniclabs/carmen/blob/1af2f6427380d31c0e0fe136cdee756a5bbf90db/go/common/update.go#L44) field. Would it be sufficient to set the code for these accounts to the empty code instead of tracking the "maybeEmptyCodeAccounts" here?

go/database/vt/memory/state_test.go
```diff
@@ -466,6 +465,36 @@ func TestState_IncrementalStateUpdatesResultInSameCommitments(t *testing.T) {
 	}
 }
 
+func TestState_EoAccount_CodeHash_Is_Eth_Empty_Hash(t *testing.T) {
```
HerbertJordan, line 468: I do not understand this test case name ...

go/database/vt/memory/state.go
kjezek: ok, conigured

Issue #62, opened 2025-07-30
Clear up meaning of the word "state" in FFI interface


psalz:
From discussion with @LorenzSchueler

The current naming scheme of the FFI interface is very confusing:

- The term "state" is used to describe both the Live and Archive DB. This is the opaque `state` parameter that can be passed into most functions (e.g. `GetBalance`, `GetNonce`, ...).
- The `OpenState` function receives a [`StateImpl`](https://github.com/0xsoniclabs/carmen/blob/da1bdc1b27bc81a4526f23dca6b5cff12b78ad49/cpp/state/c_state.h#L50) and `ArchiveImpl` parameter, the first should be called `LiveImpl`.
- `GetAccountState` uses the term in a completely different manner; here it refers to whether an account exists or not. This should be renamed to something like `AccountExists`.

Earlier suggestion by @HerbertJordan 

> rename `OpenState` to `OpenDatabase` and return a new `C_Database` type
> update update Flush and Close to take a `C_Database`
> add a new `ReleaseDatabase` function
> add a new `C_State GetLiveState(C_Database)` function to get access to the mutable head-state of the database

Alternative suggestion by @LorenzSchueler 

- Remove `GetArchiveState` completely
- Pass block height into all getter functions. Internally these will then dispatch either to the LiveDB or ArchiveDB.

psalz:
Fixed via #66, #69 and #70.

PR #69, opened 2025-08-01
Split FFI interface into Database and State interface

LorenzSchueler:
This is the second part of the FFI refactor (see #62).

Currently, the FFI interface uses a single type (alias) called `C_State` which is used for operations that semantically operate on a database and also ones that operate on a state (live state or archive state at a certain block height).
The state that gets returned from `OpenState` represents both the database and the live state. To get an archive state `GetArchiveState` is used which is a bit weird considering that the `state` object which gets passed in is a live state.

This PR splits the FFI interface into a Database and a State interface.
While we can not really enforce types in an FFI interface (and most type aliases are aliases to `void` anyway), we can at least convey the semantics by having different type aliases and function names.
Also, `OpenDatabase` is no longer a handle to a live state. Instead, the live state is retrieved by calling `GetLiveState`.

- add `C_Database` type alias
- rename `OpenState` to `OpenDatabase`
- add `ReleaseDatabase` function
- `Flush`, `Close`, `GetMemoryFootprint` and `GetArchiveState` now take a `C_Database` instead of `C_State` as argument
- add `GetLiveState`

### DIFF
--- a/rust/src/ffi/exported.rs
+++ b/rust/src/ffi/exported.rs
@@ -783,13 +783,13 @@ unsafe extern "C" fn Carmen_Rust_GetHash(state: *mut c_void, out_hash: *mut c_vo
 /// caller.
 ///
 /// # Safety
-/// - `state` must be a valid pointer to a `StateWrapper` object which holds a pointer to a `dyn
-///   CarmenState` object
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Looks good, thanks!

PR #72, opened 2025-08-11
Add Node, NodeId type and Commitment trait

LorenzSchueler:
This PR adds the node ID type, multiple node types and the commitment trait.

Currently, there are 4 node types which are wrapped in the `Node` enum.

There are two types of leaf nodes, one for 2 and one for 256 children. The one for 2 children uses the generic struct `SparseLeafNode`. Using this type, we can easily add more sizes later on if performance results suggests that this improves performance. For inner nodes there is currently only the full variant with 256 children, however we can also later one add a sparse variant.

To avoid wasting memory, the enum `Node` holds boxed inner and leaf nodes. This way the `Node` is only two bytes large, and the allocation of box depends on the size of the inner or leaf node. Without the boxing, all nodes would be as large as the largest variant, which would mean that the leaf node with 2 children would need 128x more memory than with the current implementation.

The `NodeId` will only be interacted with in the file layer. All other layers just treat it as an opaque identifier. The ID is 6 bytes large and uses the upper bits to encode the node type.

The commitment is currently a type without any methods. Those will be filled when needed. `()` implements `Commitment`, with the intention that we use `()` when we do not want to store a commitment in the storage backend. For the case that we want to store the commitment in the nodes, we simply add that type and implement `Commitment` for it.

### DIFF
--- a/rust/src/types/id.rs
+++ b/rust/src/types/id.rs
@@ -12,7 +12,7 @@ use zerocopy::{FromBytes, Immutable, IntoBytes, Unaligned};
 
 use crate::types::NodeType;
 
-/// An identifier for a node in a file based Verkle trie.
+/// An identifier for a node in a (file-based) Verkle trie.
 // NOTE: Changing the layout of this struct will break backwards compatibility of the
 // serialization format.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, FromBytes, IntoBytes, Immutable, Unaligned)]
@@ -20,17 +20,22 @@ use crate::types::NodeType;
 pub struct NodeId([u8; 6]);
 
 impl NodeId {
-    pub const EMPTY_NODE_PREFIX: u64 = 0x0000_0000_0000_0000;
-    pub const INNER_NODE_PREFIX: u64 = 0x0000_4000_0000_0000;
-    pub const LEAF_NODE_2_PREFIX: u64 = 0x0000_8000_0000_0000;
-    pub const LEAF_NODE_256_PREFIX: u64 = 0x0000_C000_0000_0000;
+    // The upper 2 bits are used to encode the node type.
+    const EMPTY_NODE_PREFIX: u64 = 0x0000_0000_0000_0000;
+    const INNER_NODE_PREFIX: u64 = 0x0000_4000_0000_0000;
+    const LEAF_NODE_2_PREFIX: u64 = 0x0000_8000_0000_0000;
+    const LEAF_NODE_256_PREFIX: u64 = 0x0000_C000_0000_0000;
 
     const PREFIX_MASK: u64 = 0x0000_C000_0000_0000;
-    const INDEX_MASK: u64 = !Self::PREFIX_MASK;
+    const INDEX_MASK: u64 = 0x0000_3FFF_FFFF_FFFF;
 
     /// Creates a new [`NodeId`] from a [`u64`] index and a [`NodeType`].
     /// The index must be smaller than 2^46.
     pub fn from_idx_and_node_type(idx: u64, node_type: NodeType) -> Self {
+        assert!(
+            (idx & !Self::INDEX_MASK) == 0,
+            "indices can not get this large, unless we have a bug somewhere"
+        );
         let prefix = match node_type {
             NodeType::Empty => Self::EMPTY_NODE_PREFIX,
             NodeType::Inner => Self::INNER_NODE_PREFIX,
@@ -55,7 +60,7 @@ impl NodeId {
             Self::LEAF_NODE_256_PREFIX => Some(NodeType::Leaf256),
             // There are only two ways to create a NodeId:
             // - Using `from_idx_and_node_type` with guarantees that the prefix is valid.
-            // - Deserializing a `SerializedId` which may hold invalid prefixes in case the data was
+            // - Deserializing from a file which may hold invalid prefixes in case the data was
             //   corrupted.
             _ => None,
         }
@@ -80,8 +85,7 @@ mod tests {
 
     #[test]
     fn from_idx_and_node_type_creates_id_from_lower_6_bytes_logic_or_node_type_prefix() {
-        let idx = 0x1234_5678_90ab_cdef;
-        let lower_6_bytes_idx = 0x5678_90ab_cdef;
+        let idx = 0x0000_1234_5678_9abc;
         let cases = [
             (NodeType::Empty, 0x0000_0000_0000_0000),
             (NodeType::Inner, 0x0000_4000_0000_0000),
@@ -91,8 +95,16 @@ mod tests {
 
         for (node_type, prefix) in cases {
             let id = NodeId::from_idx_and_node_type(idx, node_type);
-            assert_eq!(id.to_u64(), lower_6_bytes_idx | prefix);
+            assert_eq!(id.to_u64(), idx | prefix);
         }
+    }
+
+    #[test]
+    #[should_panic]
+    fn from_idx_and_node_type_panics_if_index_too_large() {
+        let idx = 0x0000_f000_0000_0000;
+
+        NodeId::from_idx_and_node_type(idx, NodeType::Empty);
     }
 
     #[test]
--- a/rust/src/types/node.rs
+++ b/rust/src/types/node.rs
@@ -12,7 +12,7 @@ use zerocopy::{FromBytes, Immutable, IntoBytes, Unaligned};
 
 use crate::types::{Commitment, NodeId, Value};
 
-/// A value of a leaf node in a (file based) Verkle trie with its index.
+/// A value of a leaf node in a (file-based) Verkle trie, together with its index.
 // NOTE: Changing the layout of this struct will break backwards compatibility of the
 // serialization format.
 #[derive(
@@ -26,7 +26,7 @@ pub struct ValueWithIndex {
     pub value: Value,
 }
 
-/// A sparsely populated leaf node in a (file based) Verkle trie.
+/// A sparsely populated leaf node in a (file-based) Verkle trie.
 // NOTE: This type should NOT implement [`Clone`] because there should never be two instances
 // corresponding to the same logical node.
 // NOTE: Changing the layout of this struct will break backwards compatibility of the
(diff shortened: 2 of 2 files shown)


rust/src/types/commitment.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+/// A Verkle trie node commitment.
+pub trait Commitment {}
+
+/// In case we don't store an commitment, we use the unit type.
```
psalz, line 14:
```suggestion
/// In case we don't store a commitment, we use the unit type.
```

Is this a use case we discussed? When would you want to not store a commitment?

rust/src/types/node.rs
```diff
@@ -0,0 +1,180 @@
...
+    Empty,
+    Inner(Box<InnerNode<C>>),
+    Leaf2(Box<SparseLeafNode<2, C>>),
+    Leaf256(Box<FullLeafNode<C>>),
+}
+
+/// A node type of a node in a (file based) Verkle trie.
+/// This type is primarily used for conversion between [`Node`] and indexes in the file storage.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NodeType {
```
ihtefyw, line 128: I don't like duplicating the enum but I suppose there is no other way to reserve IDs. I would like to have a programmatic way of enforcing a sync between the two enums. 
Maybe a From between the two?  
WDYT? Or can we assume they will be kept in sync as they are so closely connected?

ihtefyw (approved):

rust/src/types/id.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub const INNER_NODE_PREFIX: u64 = 0x0000_4000_0000_0000;
+    pub const LEAF_NODE_2_PREFIX: u64 = 0x0000_8000_0000_0000;
+    pub const LEAF_NODE_256_PREFIX: u64 = 0x0000_C000_0000_0000;
+
+    const PREFIX_MASK: u64 = 0x0000_C000_0000_0000;
+    const INDEX_MASK: u64 = !Self::PREFIX_MASK;
+
+    /// Creates a new [`NodeId`] from a [`u64`] index and a [`NodeType`].
+    /// The index must be smaller than 2^46.
+    pub fn from_idx_and_node_type(idx: u64, node_type: NodeType) -> Self {
```
psalz, line 33:
```suggestion
    pub fn from_idx_and_node_type(idx: u64, node_type: NodeType) -> Self {
        assert!(idx < 2^46);
```

Why not check for this precondition?

rust/src/types/id.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub fn to_node_type(self) -> Option<NodeType> {
+        match self.to_u64() & Self::PREFIX_MASK {
+            Self::EMPTY_NODE_PREFIX => Some(NodeType::Empty),
+            Self::INNER_NODE_PREFIX => Some(NodeType::Inner),
+            Self::LEAF_NODE_2_PREFIX => Some(NodeType::Leaf2),
+            Self::LEAF_NODE_256_PREFIX => Some(NodeType::Leaf256),
+            // There are only two ways to create a NodeId:
+            // - Using `from_idx_and_node_type` with guarantees that the prefix is valid.
+            // - Deserializing a `SerializedId` which may hold invalid prefixes in case the data was
+            //   corrupted.
```
psalz, line 59:
```suggestion
            // - Deserializing from a file (not yet implemented), which may hold invalid prefixes in case the data was
            //   corrupted.
```

rust/src/types/node.rs
LorenzSchueler: Yeah, I'm also not too happy with them. They are not even used to reserve IDs directly because there we still pass a reference to a `Node`.
They are used in the `FileStoreManager` (future PR) like so:
```rs
    fn get(&self, id: NodeId) -> Result<Node<C>, Error> {
        let idx = id.to_index();
        match id.to_node_type().ok_or(Error::InvalidId)? {
            NodeType::Empty => Ok(Node::Empty),
            NodeType::Inner => {
                let node = self.inner_nodes.get(idx)?;
                Ok(Node::Inner(Box::new(node)))
            }
            NodeType::Leaf2 => {
                let node = self.leaf_nodes_2.get(idx)?;
                Ok(Node::Leaf2(Box::new(node)))
            }
            NodeType::Leaf256 => {
                let node = self.leaf_nodes_256.get(idx)?;
                Ok(Node::Leaf256(Box::new(node)))
            }
        }
    }
    
    fn delete(&self, id: NodeId) -> Result<(), Error> {
        let idx = id.to_index();
        match id.to_node_type().ok_or(Error::InvalidId)? {
            NodeType::Empty => Ok(()),
            NodeType::Inner => self.inner_nodes.delete(idx),
            NodeType::Leaf2 => self.leaf_nodes_2.delete(idx),
            NodeType::Leaf256 => self.leaf_nodes_256.delete(idx),
        }
    }
```

The other option would be to return the prefix and do the match on the prefix here instead of in `Id::to_node_type`
```
    pub fn to_node_type(self) -> Option<NodeType> {
        match self.to_u64() & Self::PREFIX_MASK {
            Self::EMPTY_NODE_PREFIX => Some(NodeType::Empty),
            Self::INNER_NODE_PREFIX => Some(NodeType::Inner),
            Self::LEAF_NODE_2_PREFIX => Some(NodeType::Leaf2),
            Self::LEAF_NODE_256_PREFIX => Some(NodeType::Leaf256),
            // There are only two ways to create a NodeId:
            // - Using `from_idx_and_node_type` with guarantees that the prefix is valid.
            // - Deserializing a `SerializedId` which may hold invalid prefixes in case the data was
            //   corrupted.
            _ => None,
        }
    }
```

But this is even easier to get out of sync.

rust/src/types/id.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// An identifier for a node in a file based Verkle trie.
+// NOTE: Changing the layout of this struct will break backwards compatibility of the
+// serialization format.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, FromBytes, IntoBytes, Immutable, Unaligned)]
+#[repr(transparent)]
+pub struct NodeId([u8; 6]);
+
+impl NodeId {
+    pub const EMPTY_NODE_PREFIX: u64 = 0x0000_0000_0000_0000;
```
psalz, line 23:
```suggestion
    // We use the upper 2 bits to encode the node type.
    pub const EMPTY_NODE_PREFIX: u64 = 0x0000_0000_0000_0000;
```

rust/src/types/id.rs
```diff
@@ -0,0 +1,145 @@
...
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+use zerocopy::{FromBytes, Immutable, IntoBytes, Unaligned};
+
+use crate::types::NodeType;
+
+/// An identifier for a node in a file based Verkle trie.
```
psalz, line 15:
```suggestion
/// An identifier for a node in a file-based Verkle trie.
```

Here and in a couple more places below.

rust/src/types/node.rs
```diff
@@ -0,0 +1,180 @@
...
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+use zerocopy::{FromBytes, Immutable, IntoBytes, Unaligned};
+
+use crate::types::{Commitment, NodeId, Value};
+
+/// A value of a leaf node in a (file based) Verkle trie with its index.
```
psalz, line 15:
```suggestion
/// A value of a leaf node in a (file based) Verkle trie, together with its index.
/// Used to identify values in sparsely populated leaf nodes.
```

rust/src/types/node.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+impl<const N: usize, C> Default for SparseLeafNode<N, C>
+where
+    C: Commitment + Default,
+{
+    fn default() -> Self {
+        let mut values = [ValueWithIndex::default(); N];
+        values.iter_mut().enumerate().for_each(|(i, v)| {
+            v.index = i as u8;
```
psalz, line 50: Is there really a point to doing this? We'll overwrite this in most cases anyway, no?

rust/src/types/node.rs
```diff
@@ -0,0 +1,180 @@
...
+    Leaf256,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn sparse_leaf_node_default_returns_leaf_node_with_all_values_set_to_default_and_unique_indices()
+     {
```
psalz, line 141: Uhm, why was this not caught by CI? Line too long?

rust/src/types/node.rs
```diff
@@ -0,0 +1,180 @@
...
+{
+    fn default() -> Self {
+        InnerNode {
+            commitment: C::default(),
+            values: [NodeId::from_idx_and_node_type(0, NodeType::Empty); 256],
+        }
+    }
+}
+
+/// A node in a (file based) Verkle trie.
```
psalz, line 113:
```suggestion
/// A node in a (file based) Verkle trie.
/// Non-empty nodes are stored as boxed to save on memory (otherwise the size of [Node] would be dictated by the largest variant).
```

rust/src/types/node.rs
psalz: Why can't you just use a C-style enum?
```rust
#[repr(u64)]
enum NodeType  {
    Empty = 0x0000_0000_0000_0000,
    Inner = 0x0000_4000_0000_0000,
    Leaf2 = 0x0000_8000_0000_0000,
    Leaf256 = 0x0000_C000_0000_0000,
}
```

rust/src/types/commitment.rs
LorenzSchueler: As discusses, we don't need the trait as we will be using always the same commitment type.

rust/src/types/node.rs
LorenzSchueler: What should be caught? The line is auto formatted.

rust/src/types/node.rs
psalz: Hm ok, I still find it somewhat strange that in this hypothetical scenario you might end up with a leaf that contains values `[(42, 0x123), (1, 0x0)]` though.

PR #75, opened 2025-08-12
Add FileBackend trait and pread&pwrite and seeking based impls

LorenzSchueler:
This PR adds the `FileBackend` trait and adds two implementations, one using [pread & pwrite](https://www.man7.org/linux/man-pages/man2/pread.2.html) and one using seeking operations for non-Unix platforms.
Later on, we might add more implementations.

The `FileBackend` is the lowest layer in the file based storage backend (the blue parts).
Carmen Layers

New dev-dependencies:
- [tempfile](https://crates.io/crates/tempfile) a widely used crate to create temporary directories and delete then on drop. This is used a lot in tests.

Depends on #81 #82

### DIFF
--- a/rust/src/storage/file/file_backend.rs
+++ b/rust/src/storage/file/file_backend.rs
@@ -1,0 +1,484 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+use std::{
+    fs::OpenOptions,
+    io::{Read, Seek, SeekFrom, Write},
+    os::unix::fs::FileExt,
+    path::Path,
+    sync::Mutex,
+};
+
+/// An abstraction for concurrent file operations.
+///
+/// Implementations of this trait are required to ensure that concurrent operations of the
+/// implementing type are safe (in that there are no data races) as long as the operations operate
+/// on non-overlapping regions of the file.
+/// When called with overlapping regions, the behavior is undefined and may lead to data corruption.
+#[cfg_attr(test, mockall::automock)]
+pub trait FileBackend {
+    /// Opens a file at the given path with the specified options and tries to acquire an advisory
+    /// file lock.
+    fn open(path: &Path, options: &OpenOptions) -> std::io::Result<Self>
+    where
+        Self: Sized;
+
+    /// Writes the entire content of `buf` starting at the given `offset`.
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> std::io::Result<()>;
+
+    /// Fills the entire `buf` with data read from the file starting at the given `offset`.
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> std::io::Result<()>;
+
+    /// Flushes all changes to disk.
+    fn flush(&self) -> std::io::Result<()>;
+
+    /// Returns the length of the file.
+    fn len(&self) -> Result<u64, std::io::Error>;
+
+    /// Sets the length of the file to `len`.
+    /// This may truncate the file if `len` is smaller than the current length.
+    fn set_len(&self, len: u64) -> std::io::Result<()>;
+}
+
+/// A wrapper around [`std::fs::File`] that implements [`FileBackend`] using a mutex to ensure
+/// exclusive access to the file. This is suitable for platforms where `pread` and `pwrite` are not
+/// available or where seeking is required for other reasons.
+pub struct SeekFile(Mutex<std::fs::File>);
+
+impl FileBackend for SeekFile {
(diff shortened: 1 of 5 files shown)


rust/src/storage/file/file_backend.rs
```diff
@@ -0,0 +1,484 @@
...
+    fn set_len(&self, len: u64) -> std::io::Result<()> {
+        self.0.lock().unwrap().set_len(len)
+    }
+}
+
+/// A wrapper around [`std::fs::File`] that implements [`FileBackend`] using the Unix-specific file
+/// operations `pread` and `pwrite` which do not modify the file offset. This avoids the syscall for
+/// seeking and allows for concurrent access without needing to manage a cursor.
+#[cfg(unix)]
+pub struct NoSeekFile(std::fs::File);
```
LuisPH3, line 94: please avoid tuple types. 
they hurt readability. a lot

rust/src/storage/file/file_backend.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    /// Flushes all changes to disk.
+    fn flush(&self) -> std::io::Result<()>;
+
+    /// Returns the length of the file.
+    fn len(&self) -> Result<u64, std::io::Error>;
+
+    /// Sets the length of the file to `len`.
+    /// This may truncate the file if `len` is smaller than the current length.
+    fn set_len(&self, len: u64) -> std::io::Result<()>;
```
LuisPH3, line 47: maybe call it truncate. files do not have lenght but size.

rust/src/storage/file/file_backend.rs
```diff
@@ -0,0 +1,484 @@
...
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> std::io::Result<()>;
+
+    /// Fills the entire `buf` with data read from the file starting at the given `offset`.
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> std::io::Result<()>;
+
+    /// Flushes all changes to disk.
+    fn flush(&self) -> std::io::Result<()>;
+
+    /// Returns the length of the file.
+    fn len(&self) -> Result<u64, std::io::Error>;
```
LuisPH3, line 43: size. files do not have lenght, but size.
please specify unit. Bytes I pressume.

rust/src/storage/file/file_backend.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// file lock.
+    fn open(path: &Path, options: &OpenOptions) -> std::io::Result<Self>
+    where
+        Self: Sized;
+
+    /// Writes the entire content of `buf` starting at the given `offset`.
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> std::io::Result<()>;
+
+    /// Fills the entire `buf` with data read from the file starting at the given `offset`.
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> std::io::Result<()>;
```
LuisPH3, line 37: if there is no other method called read, why do you want to call this with such a long name?

LuisPH3 (approved):
Some notes about naming.

rust/src/storage/file/file_backend.rs
LorenzSchueler: In general (when storing multiple fields) I completely agree. But in the case of the new type pattern, where only a single field is stored, I'm not sure if `file.0` is worse than `file.file` or `file.inner` or something like that.

rust/src/storage/file/file_backend.rs
LorenzSchueler: https://doc.rust-lang.org/std/fs/struct.File.html#method.set_len
std also calls it `set_len`

rust/src/storage/file/file_backend.rs
```diff
@@ -0,0 +1,484 @@
...
+#[cfg_attr(test, mockall::automock)]
+pub trait FileBackend {
+    /// Opens a file at the given path with the specified options and tries to acquire an advisory
+    /// file lock.
+    fn open(path: &Path, options: &OpenOptions) -> std::io::Result<Self>
+    where
+        Self: Sized;
+
+    /// Writes the entire content of `buf` starting at the given `offset`.
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> std::io::Result<()>;
```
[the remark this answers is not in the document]
LorenzSchueler: The "all" means that the full `buf` is guaranteed to be written out if the function returns successfully. The naming corresponds to https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html#method.write_all_at

rust/src/storage/file/file_backend.rs
LuisPH3: This trait contains other methods: `read_at, write_at ` this explains the very convoluted naming. 

I would argue that an abstraction needs to abstract something.

rust/src/storage/file/file_backend.rs
LorenzSchueler: I thought you are talking about the function name. I'll update the doc string.

psalz (approved):

PR #76, opened 2025-08-12
makes diagnostics tools generally available

kjezek:
This PR moves original `addPerformanceDiagnoses` to its own package and makes it public,

it is ment to be usable for tools stored in different packages

### DIFF
--- a/go/common/diagnostics/diagnostics.go
+++ b/go/common/diagnostics/diagnostics.go
@@ -1,0 +1,104 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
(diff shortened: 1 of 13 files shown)


go/common/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func startDiagnosticServer(port int) {
+	if port <= 0 || port >= (1<<16) {
+		return
+	}
+	fmt.Printf("Starting diagnostic server at port http://localhost:%d\n", port)
+	fmt.Printf("(see https://pkg.go.dev/net/http/pprof#hdr-Usage_examples for usage examples)\n")
+	fmt.Printf("Block and mutex sampling rate is set to 100%% for diagnostics, which may impact overall performance\n")
```
cabrador, line 67:
```suggestion
	fmt.Printf("Block and mutex sampling rate is set to 100% for diagnostics, which may impact overall performance\n")
```

cabrador (approved):

go/common/diagnostics/diagnostics.go
kjezek: the percentage sign has to be doubled not to interpret is as an escape sequence

PR #77, opened 2025-08-13
generalise insertion benchmark

kjezek:
This PR moves MPT benchmark to State. 

It works against the state, and for this reason, it is more fitting here. It also enables option to run benchmark against other schemas, not only MPT.


go/database/mpt/tool/stress.go
```diff
@@ -370,3 +381,18 @@ func getMemoryUsage() uint64 {
 	runtime.ReadMemStats(&m)
 	return m.Alloc
 }
+
+// GetDirectorySize computes the size of all files in the given directory in bytes.
```
cabrador, line 385: nit: lowercase

cabrador (approved):

go/database/mpt/tool/stress.go
kjezek: right, fixed

cabrador (approved):

cabrador (dismissed):

PR #87, opened 2025-08-14
Add `ErrorState` type

ihtefyw:
> This is in preparation of the NodeCache 

This PR adds the `ErrorState` type, a thread-safe utility to store an error.
The main reason for this is that we need a way to store errors happening in worker threads during a request in order to return them through the FFI interface (see #79). 
The usage would be:
- Instantiate an `ErrorState` in a `CarmenState` request
- Propagate it through the calls to the Trie, Cache, and Storage
- Check the `ErrorState` object and return the error.

This is also used in the subsequent `NodeCache` PRs to store storage errors happening during cache eviction due to a limitation of the `quickcache` interface.

### DIFF
--- a/rust/src/error.rs
+++ b/rust/src/error.rs
@@ -8,6 +8,9 @@
 // On the date above, in accordance with the Business Source License, use of
 // this software will be governed by the GNU Lesser General Public License v3.
 
+#[cfg(test)]
+use std::{ops::Deref, sync::Mutex};
+
 use thiserror::Error;
 
 use crate::storage;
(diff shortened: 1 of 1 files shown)


rust/src/error.rs
```diff
@@ -25,3 +28,68 @@ pub enum Error {
...
+
+        // Registering another error should not change the existing one
+        state.register(Error::UnsupportedOperation("test".to_string()));
+        {
+            let guard = state.get();
+            assert!(guard.is_some());
+            assert!(matches!(
+                guard.as_ref().unwrap(),
+                Error::UnsupportedSchema(1)
+            ));
```
LorenzSchueler, line 92: nit: WDYT?
```suggestion
            assert!(matches!(*guard, Some(Error::UnsupportedSchema(1))));
```

LorenzSchueler (changes requested):
I think it should do what we need 👍🏽

rust/src/error.rs
```diff
@@ -25,3 +28,62 @@ pub enum Error {
...
+        assert!(state.get().is_none());
+    }
+
+    #[test]
+    fn error_state_register_sets_error_if_none_exists() {
+        let state = ErrorState::new();
+        state.store(Error::UnsupportedSchema(1));
+        {
+            let err = state.get();
+            assert!(err.is_some());
```
LorenzSchueler, line 77: This is asserted in the next line already
```suggestion
```

rust/src/error.rs
```diff
@@ -25,3 +28,62 @@ pub enum Error {
...
+            assert!(matches!(*err, Some(Error::UnsupportedSchema(1))));
+        }
+
+        // Registering another error should not change the existing one
+        state.store(Error::UnsupportedOperation("test".to_string()));
+        {
+            let err = state.get();
+            assert!(err.is_some());
```
LorenzSchueler, line 85: This is asserted in the next line already
```suggestion
```

LorenzSchueler (dismissed):

LorenzSchueler (dismissed):

LuisPH3 (changes requested):
IMO, this infrastructure is dangerous.
It is used for background threads to accumulate errors, allowing future calls, if any, to forward them to the software using the library.
I think it would be much better not to merge this now, just log or panic at the location where this utility is used and instead introduce tooling later during development. where the use of this tool is visible and can be discussed.

I would suggest not using this tool now, focus on MVP and deal with async errors once the system is running.

PR #90, opened 2025-08-18
Handle error codes returned via FFI in Go

LorenzSchueler:
This PR uses the error codes introduced in #79 to return errors on the Go side if an operation failed.

The only exception is if `C.Carmen_Cpp_GetMemoryFootprint` fails. Unlike all other methods, `State::GetMemoryFootprint` does not return an error. One option would be to change the interface, but this requires changes in a lot of different places in the codebase and would also require either changing other interfaces as well because they now also need to be able to return an error, or swallow the error silently. The other option (which is implemented in this PR), is to set the note on the memory footprint object to an appropriate error message and log the error.

Closes https://github.com/0xsoniclabs/sonic-admin/issues/322

### DIFF
--- a/go/state/cppstate/cpp_state.go
+++ b/go/state/cppstate/cpp_state.go
@@ -25,6 +25,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


psalz (dismissed):

ihtefyw (approved):

go/state/cppstate/cpp_state.go
```diff
@@ -72,16 +73,16 @@ func newState(impl C.enum_LiveImpl, params state.Parameters) (state.State, error
 	}
 
 	db := unsafe.Pointer(nil)
-	C.Carmen_Cpp_OpenDatabase(C.C_Schema(params.Schema), impl, C.enum_ArchiveImpl(archive), dir, C.int(len(params.Directory)), &db)
-	if db == unsafe.Pointer(nil) {
-		return nil, fmt.Errorf("%w: failed to create C++ database instance for parameters %v", state.UnsupportedConfiguration, params)
+	result := C.Carmen_Cpp_OpenDatabase(C.C_Schema(params.Schema), impl, C.enum_ArchiveImpl(archive), dir, C.int(len(params.Directory)), &db)
+	if result != C.kResult_Success || db == unsafe.Pointer(nil) {
```
kjezek, line 77: I think here and below, the error message covers two issues: either unsupported configuration, or an error state. It may be confusing, because there are two distinct situations. Either the configuration is not supported, or opening a database caused an error.  I think it should produce two distinct error messages based on what has really happened.

kjezek (changes requested):
looks good thanks, I have just one concerns, see my comment in the code.

It seems that one error message covers two situations, either unsupported configuration, or an error. I guess it should be split in two error messages.

PR #93, opened 2025-08-19
Add StorageWithEvictionCache

LorenzSchueler:
This PR adds the `StorageWithEvictionCache` which is the layer above the `FileStoreManager`. It implements the `Storage` trait added in #74 and holds an instance of the underlying storage layer, the eviction cache and a thread pool of eviction workers.

Calls to `reserve` are just forwarded to the underlying storage.

`get` first checks if elements are in the eviction cache. If they are in the cache and have been deleted, *not found* is returned. If they are in the case and have been modified, they are removed from the cache and returned. If they are not found in the cache, the underlying storage is queried.

`set` and `delete` add the entries to the eviction cache.

The eviction workers are responsible for asynchronously writing out deletions and updates to the underlying storage.
In the case of updates, they lock the node, call the storage then remove it from the eviction cache and mark it as clean again (for the case that the node was queried in the meantime and is now again in the node cache).
Deletions are forwarded to the storage and afterward removed from the eviction cache.

To make sure that nodes can only be marked as clean by this layer, and nowhere else in the codebase, `CachedNode::set_clean` takes a `WriteCertificate`. This is a public type which can only be constructed in `storage_with_eviction_cache`.

Carmen Layers

Note: I did not add any tests for `CachedNode::set_clean`, because `WriteCertificate` can not be constructed in `cache.rs` and the dirty field can not be accessed in `storage_with_evicion_cache.rs`. But since the function has only 2 lines, I guess this is ok.

New dependencies:
- [dashmap](https://crates.io/crates/dashmap) a concurrent hash map used for the eviction cache.

Closes https://github.com/0xsoniclabs/sonic-admin/issues/274


rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+    fn delete(&self, id: NodeId) -> Result<(), Error> {
+        self.cache.insert(id, Op::Delete);
+        Ok(())
+    }
+
+    fn flush(&self) -> Result<(), Error> {
+        // busy loop until all eviction workers are done
+        // because there are no concurrent inserts, len() might only return a number that is higher
+        // that the actual number of items. This is however not a problem because we will wait a
+        // little bit longer
```
ihtefyw, line 115: old comment?

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+    pub fn shutdown(self) -> Result<(), Error> {
+        self.shutdown.store(true, Ordering::SeqCst);
+        for worker in self.workers {
+            worker.join().unwrap()?;
+        }
+        Ok(())
+    }
+}
+
+type EvictionCache = DashMap<NodeId, Op>;
```
ihtefyw, line 209: I would move  the type definition up in the file, you use it in the cache impl but it's defined way later

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+        // busy loop until all eviction workers are done
+        // because there are no concurrent inserts, len() might only return a number that is higher
+        // that the actual number of items. This is however not a problem because we will wait a
+        // little bit longer
+        while !self.cache.is_empty() {}
+        self.storage.flush()
+    }
+}
+
+struct EvictionWorkers {
```
ihtefyw, line 121: Missing docstring

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+    workers: Vec<std::thread::JoinHandle<Result<(), Error>>>,
+    shutdown: Arc<AtomicBool>, // Arc for shared ownership with eviction worker threads
+}
+
+impl EvictionWorkers {
+    const WORKER_COUNT: usize = 10;
```
ihtefyw, line 127: Maybe add a comment that this requires benchmarks.

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+
+    pub fn new<S>(eviction_cache: &Arc<EvictionCache>, storage: &Arc<S>) -> Self
```
ihtefyw, line 129: Missing docstring

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+                std::thread::spawn(move || {
+                    EvictionWorkers::task(&eviction_cache, &*storage, &shutdown)
+                })
+            })
+            .collect();
+
+        EvictionWorkers { workers, shutdown }
+    }
+
+    fn task<S>(
```
ihtefyw, line 148: Missing docstring

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+        S: Storage<Id = NodeId, Item = Node> + Send + Sync + 'static,
+    {
+        let shutdown = Arc::new(AtomicBool::new(false));
+        let workers = (0..Self::WORKER_COUNT)
+            .map(|_| {
+                let eviction_cache = eviction_cache.clone();
+                let storage = storage.clone();
+                let shutdown = shutdown.clone();
+                std::thread::spawn(move || {
+                    EvictionWorkers::task(&eviction_cache, &*storage, &shutdown)
```
ihtefyw, line 140: What is the `&*` syntax for? In my understanding, it should be a no-op as you are creating a reference and dereferencing it immediately. Am I missing something?

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+            cache: Arc::new(DashMap::new()),
+            storage: Arc::new(MockFileStorageManager::<MockFileBackend>::new()),
+            eviction_workers: EvictionWorkers {
+                workers: Vec::new(),
+                shutdown: Arc::new(AtomicBool::new(false)),
+            },
+        };
+
+        let id = NodeId::from_idx_and_node_type(0, NodeType::Inner);
+        let node = Arc::new(RwLock::new(CachedNode::new_clean(Node::Inner(
```
ihtefyw, line 294:
```suggestion
        let node = CacheEntry::new(RwLock::new(CachedNode::new_clean(Node::Inner(
```

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+        let reserved_id = storage_with_eviction_cache
+            .reserve(&Arc::new(RwLock::new(CachedNode::new_clean(node))));
+        assert_eq!(reserved_id, id);
+        assert!(storage_with_eviction_cache.cache.get(&id).is_none());
+    }
+
+    #[test]
+    fn set_inserts_node_into_cache() {
+        let id = NodeId::from_idx_and_node_type(0, NodeType::Inner);
+        let node = Arc::new(RwLock::new(CachedNode::new_clean(Node::Inner(
```
ihtefyw, line 382:
```suggestion
        let node = CacheEntry::new(RwLock::new(CachedNode::new_clean(Node::Inner(
```

rust/src/storage/file/storage_with_eviction_cache.rs
```diff
@@ -0,0 +1,577 @@
...
+        };
+
+        let reserved_id = storage_with_eviction_cache
+            .reserve(&Arc::new(RwLock::new(CachedNode::new_clean(node))));
+        assert_eq!(reserved_id, id);
+        assert!(storage_with_eviction_cache.cache.get(&id).is_none());
+    }
+
+    #[test]
+    fn set_inserts_node_into_cache() {
```
ihtefyw, line 380:
```suggestion
    fn set_inserts_set_op_into_cache() {
```
You use this naming for the delete op, let's unify them

ihtefyw (changes requested):
Looks good overall, some style and doc issues

rust/src/storage/file/storage_with_eviction_cache.rs
LorenzSchueler: `storage` is an `Arc<S>` but we need to pass `&S`, so we dereference the Arc to get S and then take a reference to the S. Usually this is not required and a single `&` takes are of it because of deref coersion, but this does not work for generic code.

rust/src/storage/file/storage_with_eviction_cache.rs
LorenzSchueler: No this still holds.

ihtefyw (approved):

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+    storage::{Error, Storage},
+    types::{Node, NodeId},
+};
+
+/// A storage backend that uses a flush buffer to hold updates and deletions while they get
+/// written to the underlying storage layer in background threads.
+///
+/// The flush workers (the background threads that write the updates and deletions to the
+/// underlying storage layer) lock nodes while they are being written out to ensure that they are
+/// not concurrently modified.
```
psalz, line 31:
```suggestion
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+                .map(|entry| (*entry.key(), entry.value().clone()));
+
+            if let Some((id, op)) = item {
+                match op {
+                    Op::Set(node) => {
+                        // Acquire a write lock on the node to prevent concurrent accesses,
+                        // write it to the underlying storage layer and then remove it from the
+                        // flush buffer. If the node was queried during
+                        // this time, it may be again in the node buffer, but
+                        // it cannot be modified until the lock is released.
```
psalz, line 172:
```suggestion
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+            },
+            None => Ok(self.storage.get(id)?),
+        }
+    }
+
+    fn reserve(&self, node: &Self::Item) -> Self::Id {
+        let id = self.storage.reserve(node);
+        // The id may have been deleted in the underlying storage layer and reassigned here, but not
+        // yet removed from the flush buffer. In this case, we remove it from the flush
+        // buffer to ensure that it is no longer in the buffer.
```
psalz, line 96:
```suggestion
        // yet removed from the flush buffer (racing against flush workers). In this case, we remove it
        // from the flush buffer to ensure that the id no longer returns an [`Error::NotFound`].
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+};
+
+/// A storage backend that uses a flush buffer to hold updates and deletions while they get
+/// written to the underlying storage layer in background threads.
+///
+/// The flush workers (the background threads that write the updates and deletions to the
+/// underlying storage layer) lock nodes while they are being written out to ensure that they are
+/// not concurrently modified.
+///
+/// Deletions only involve the id, not the node itself, so they are not locked. The id is deleted in
```
psalz, line 33:
```suggestion
/// Upon deletion, the id is deleted in
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+/// deleted in the underlying storage layer and reassigned by a concurrent task before the flush
+/// worker deleted it from the flush buffer, the `StorageWithFlushBuffer:reserve` method also
+/// deletes the id from the flush buffer. This way it is guaranteed that a reassigned id is not
+/// found in the flush buffer.
+///
+/// Queries always check the flush buffer first. If the id is found there and it is an update, the
+/// node is removed from the flush buffer and returned. The node is then still considered dirty.
+/// If the id is found in the flush buffer and it is a delete operation, a not found error is
+/// returned. Only if the id is not found in the flush buffer, the underlying storage layer is
+/// queried.
```
psalz, line 44:
```suggestion
/// Queries always check the flush buffer first. If the id is found there and it is an update, a
/// copy of the node is returned, but the node is kept in the flush buffer. This ensures that
/// once changes to a node enter the flush buffer, they will eventually be persisted. If a node
/// is updated repeatedly before a flush worker manages to flush the changes to the underlying
/// storage layer, later updates replace earlier ones and only the latest changes are persisted.
/// If the id is found in the flush buffer and it is a delete operation, a not found error is
/// returned. Only if the id is not found in the flush buffer, the underlying storage layer is
/// queried.
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+enum Op {
+    Set(Node),
+    Delete,
+}
+
+/// A zero-sized type used as a certificate that a node was written to disk and can therefore be
+/// treated as clean again.
+// Note: The struct has to have a private field so that it cannot be instantiated outside of this
+// module.
+pub struct WriteCertificate(());
```
psalz, line 223:
```suggestion
```

rust/src/types/node.rs
```diff
@@ -84,8 +78,8 @@ impl Default for FullLeafNode {
 // corresponding to the same logical node.
 // NOTE: Changing the layout of this struct will break backwards compatibility of the
 // serialization format.
-#[derive(Debug, PartialEq, Eq, FromBytes, IntoBytes, Immutable, Unaligned)]
-#[cfg_attr(test, derive(Clone))] // For testing purposes only
+#[derive(Debug, PartialEq, Eq, FromBytes, IntoBytes, Immutable, Unaligned, Clone)]
```
psalz, line 81:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, FromBytes, IntoBytes, Immutable, Unaligned)]
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+
+        // this opens:
+        // StorageWithFlushBuffer
+        //   -> FileStoreManager
+        //     -> NodeFileStorage for InnerNode
+        //       -> NoSeekFile
+        //     -> NodeFileStorage for LeafNode2
+        //       -> NoSeekFile
+        //     -> NodeFileStorage for LeafNode256
+        //       -> NoSeekFile
```
psalz, line 252:
```suggestion
        //     -> A NodeFileStorage for each node type (InnerNode, SparseLeafNode<N>, ...)
        //       -> NoSeekFile
```

The different node types aren't set in stone, so we shouldn't be too specific here.

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+        assert!(!thread.is_finished());
+        std::thread::sleep(Duration::from_millis(100));
+        // flush is still waiting
+        assert!(!thread.is_finished());
+
+        // remove the item from the buffer to allow flush to complete
+        storage_with_flush_buffer.flush_buffer.remove(&id);
+
+        std::thread::sleep(Duration::from_millis(100));
+        // flush should not call flush on the underlying storage layer and return
```
psalz, line 460: I assume this should say "should call flush" and then maybe the intent was to check the mock if flush was called?

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+        };
+
+        let id = NodeId::from_idx_and_node_type(0, NodeType::Inner);
+        let node = Node::Inner(Box::default());
+
+        storage.flush_buffer.insert(id, Op::Set(node.clone()));
+
+        let result = storage.get(id).unwrap();
+        assert_eq!(result, node);
+        assert!(storage.flush_buffer.get(&id).is_some());
```
psalz, line 301:
```suggestion
        // Node is kept for eventual flush.
        assert!(storage.flush_buffer.get(&id).is_some());
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -0,0 +1,568 @@
...
+            storage.flush_workers.workers.len(),
+            FlushWorkers::WORKER_COUNT
+        );
+        for worker in &storage.flush_workers.workers {
+            assert!(!worker.is_finished()); // Ensure the worker is running
+        }
+    }
+
+    #[test]
+    fn get_returns_node_from_buffer_if_present_as_set_op() {
```
psalz, line 284:
```suggestion
    fn get_returns_copy_of_node_if_present_as_set_op() {
```

psalz (changes requested):
The behavior of this type changed quite considerably, most of the comments are outdated.

rust/src/storage/storage_with_flush_buffer.rs
ihtefyw: I think so, I think he refers to the fact that at line 460 the expectation of the mock should be fullfilled

ihtefyw (approved):

rust/src/storage/storage_with_flush_buffer.rs
psalz: Yes but will the test fail if `flush` is not being called on the mock..?

ihtefyw: Enforced with a `.times(1)` statement on the mock

psalz (approved):

ihtefyw (approved):

psalz (approved):

PR #94, opened 2025-08-19
Generalize FFI wrappers in Go to work for both C++ and Rust

psalz:
This PR prepares the Go side for calling into the Rust implementation. I've adapted the existing wrappers around the C++ FFI interface to also work for Rust. In order to not having to duplicate the entire `cpp_state.go` file, I instead opted to add one layer of indirection in the form of the `FFIInterface`, which is just a thin wrapper around the `Carmen_Cpp_*` and `Carmen_Rust_*` functions. This means that we have one additional virtual function call for each database interaction. Should this turn out the be a performance problem (unlikely), we can revisit this. Other options may include using Go generics or manually loading the libraries using `dlopen`.

I've also added the Rust build to the top-level Makefile, and changed it to build a shared library, in line with C++. It is not clear why C++ uses a shared library in the first place though, and we may want to change both to use static libraries in the future.

Note: GitHub unfortunately does't recognize that `ffi_state.go` is a rename of `ccp_state.go` and therefore shows the diff as a completely new file. You can see the changes to the file by only looking at the diff for the second commit.


go/state/ffistate/ffi_state.go
```diff
@@ -72,122 +252,137 @@ func newState(impl C.enum_LiveImpl, params state.Parameters) (state.State, error
...
 
+	var ffi FFIInterface
+
+	switch ffiType {
+	case FFITypeCpp:
+		ffi = CppInterface{}
+	case FFITypeRust:
+		ffi = RustInterface{}
+	default:
+		panic("unsupported FFI type")
```
LorenzSchueler, line 263: No sure what the policy on panicking in Go is. If we hit this case, we definitely have a bug and can not reasonably recover. The question is, if we are supposed to panic here or return an error and panic further up

LorenzSchueler (approved):
Looks Good!

go/state/ffistate/ffi_state.go
```diff
@@ -0,0 +1,631 @@
...
+const CodeMaxSize = 25000   // Contract limit is 24577
+
+type FFIType int
+
+const (
+	FFITypeCpp FFIType = iota
+	FFITypeRust
+)
+
+type FFIInterface interface {
```
HerbertJordan, line 52: Please try to eliminate implementation details like "FFI" and "Interface" from name. Focus on what this is offering. e.g "ExternDatabase" or "ExternDatabaseController";

go/state/ffistate/ffi_state_test.go
```diff
@@ -37,7 +41,11 @@ func TestAccountsAreInitiallyUnknown(t *testing.T) {
 }
 
 func TestAccountsCanBeCreated(t *testing.T) {
-	runForEachCppConfig(t, func(t *testing.T, state state.State) {
+	runForEachCppConfig(t, func(t *testing.T, state state.State, config state.Configuration) {
```
HerbertJordan, line 44: Please remove the `Cpp` in the name of this function.

HerbertJordan (dismissed):

go/state/ffistate/ffi_state_test.go
psalz: Done!

go/state/ffistate/ffi_state.go
psalz: Done!

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+const CodeMaxSize = 25000   // Contract limit is 24577
+
+type ExternalImpl int
+
+const (
+	ExternalImplCpp ExternalImpl = iota
+	ExternalImplRust
+)
+
+type ExternalBindings interface {
```
HerbertJordan, line 52: Could we make this package private?
```suggestion
type externalBindings interface {
```

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+
+const CodeCacheSize = 8_000 // ~ 200 MiB of memory for go-side code cache
+const CodeMaxSize = 25000   // Contract limit is 24577
+
+type ExternalImpl int
+
+const (
+	ExternalImplCpp ExternalImpl = iota
+	ExternalImplRust
+)
```
HerbertJordan, line 50: I guess these could and should all be not exported.

```suggestion
const codeCacheSize = 8_000 // ~ 200 MiB of memory for go-side code cache
const codeMaxSize = 25000   // Contract limit is 24577

type externalImpl int

const (
	externalImplCpp externalImpl = iota
	externalImplRust
)
```

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+	case ExternalImplRust:
+		bindings = RustBindings{}
+	default:
+		return nil, fmt.Errorf("%w: unsupported FFI type %v", state.UnsupportedConfiguration, extImpl)
+	}
+
+	db := unsafe.Pointer(nil)
+	result := bindings.OpenDatabase(C.C_Schema(params.Schema), impl, C.enum_ArchiveImpl(archive), dir, C.int(len(params.Directory)), &db)
+	if result != C.kResult_Success {
+		return nil, fmt.Errorf("failed to create FFI database instance for parameters %v (error code %v)", params, result)
```
HerbertJordan, line 269:
```suggestion
		return nil, fmt.Errorf("failed to create database instance for parameters %v (error code %v)", params, result)
```

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+	if db == unsafe.Pointer(nil) {
+		return nil, fmt.Errorf("%w: failed to create FFI database instance for parameters %v", state.UnsupportedConfiguration, params)
```
HerbertJordan, line 272:
```suggestion
		return nil, fmt.Errorf("%w: failed to create database instance for parameters %v", state.UnsupportedConfiguration, params)
```

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+	live := unsafe.Pointer(nil)
+	result = bindings.GetLiveState(db, &live)
+	if result != C.kResult_Success {
+		C.Carmen_Cpp_ReleaseDatabase(db)
```
HerbertJordan, line 278:
```suggestion
		bindings.ReleaseDatabase(db)
```

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+		return nil, fmt.Errorf("failed to create FFI live state instance for parameters %v (error code %v)", params, result)
```
HerbertJordan, line 279:
```suggestion
		return nil, fmt.Errorf("failed to create live state instance for parameters %v (error code %v)", params, result)
```

go/state/externalstate/external_state.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+	if live == unsafe.Pointer(nil) {
+		bindings.ReleaseDatabase(db)
+		return nil, fmt.Errorf("%w: failed to create FFI live state instance for parameters %v", state.UnsupportedConfiguration, params)
```
HerbertJordan, line 283:
```suggestion
		return nil, fmt.Errorf("%w: failed to create live state instance for parameters %v", state.UnsupportedConfiguration, params)
```

HerbertJordan (approved):
Looks good, just a few minor issues that should be addressed.

I think the "XxxBinding" name is fitting.

go/state/externalstate/external_state.go
psalz: Yes that makes sense (I forgot again how visibility works in Go).

ihtefyw (approved):

PR #101, opened 2025-08-28
Add `Pool` trait and remove `Cache` types

ihtefyw:
This PR adds the `Pool` trait and its basic types: `PoolItem`
The `Pool` substitutes the `Cache` component, and `PoolItem` substitutes the `Cache` types.

The `Pool` is an abstraction for a thread-safe pools of items, and it's the entry point for accessing the storage from the upcoming storage-based Verkle-trie.
The `Pool` is generic over three types:
- `Key`: The type used to index the elements in the pool
- `Type`: The type of the item handled by the pool

The `PoolItem` is the item returned from the `Pool`, defined as an `Arc<RwLock<T>>`. 
- `Arc` because the value it's shared across threads
- `RwLock` for thread-safe access

The existence of the `PoolItem` type is justified by the needing of hiding the `Arc` to avoid cloning.
The pool returns `PoolItem<impl DerefMut<Target = Type>>`. This is usually a wrapper around `Type`. This is needed as a `Pool` implementation may add metadata to `Type` for various reasons (e.g. lifecycle management, storage optimizations, etc.)

### DIFF
--- a/rust/src/pool/mod.rs
+++ b/rust/src/pool/mod.rs
@@ -22,14 +22,14 @@ pub trait Pool {
     /// The type of items indexed by the pool.
     type Type;
 
+    /// Adds the item in the pool and returns an ID for it.
+    fn add(&self, item: Self::Type) -> Result<Self::Id, Error>;
+
     /// Retrieves an item from the pool, if it exists. Returns [`Error::NotFound`] otherwise.
     fn get(
         &self,
         id: Self::Id,
     ) -> Result<PoolItem<impl DerefMut<Target = Self::Type> + Send + Sync + 'static>, Error>;
-
-    /// Stores the item in the pool and returns an ID for it.
-    fn set(&self, item: Self::Type) -> Result<Self::Id, Error>;
 
     /// Deletes an item with the given ID from the pool
     /// The ID may be reused in the future, when creating a new item by calling [`Pool::set`].
@@ -106,7 +106,7 @@ mod tests {
                 .ok_or(Error::Storage(storage::Error::NotFound))
         }
 
-        fn set(&self, value: TestNode) -> Result<Self::Id, crate::error::Error> {
+        fn add(&self, value: TestNode) -> Result<Self::Id, crate::error::Error> {
             let node = Arc::new(RwLock::new(value));
             let mut nodes = self.nodes.lock().unwrap();
             let id = nodes.len() as u32 + 1;
@@ -204,7 +204,7 @@ mod tests {
                 value: root_id * 10 + i,
                 children: vec![],
             };
-            let child_id = pool.set(child).unwrap();
+            let child_id = pool.add(child).unwrap();
             cur_node.children.push(child_id);
             let child = pool.get(child_id).unwrap();
             handles.push(thread::spawn({
@@ -299,7 +299,7 @@ mod tests {
             value: 0,
             children: vec![],
         };
-        let root_id = pool.set(root).unwrap();
+        let root_id = pool.add(root).unwrap();
 
         let root = pool.get(root_id).unwrap();
         populate_tree(&mut root.write().unwrap(), 3, TREE_DEPTH, 0, &pool);
@@ -316,7 +316,7 @@ mod tests {
             value: 0,
             children: vec![],
         };
-        let root_id = pool.set(root).unwrap();
+        let root_id = pool.add(root).unwrap();
         let root = pool.get(root_id).unwrap();
         populate_tree(&mut root.write().unwrap(), 0, TREE_DEPTH, 0, &pool.clone());
 
@@ -351,7 +351,7 @@ mod tests {
             value: 0,
             children: vec![],
         };
-        let root_id = pool.set(root).unwrap();
+        let root_id = pool.add(root).unwrap();
         let root = pool.get(root_id).unwrap();
         populate_tree(&mut root.write().unwrap(), 0, TREE_DEPTH, 0, &pool.clone());
 
@@ -376,7 +376,7 @@ mod tests {
             value: 0,
             children: vec![],
         };
-        let root_id = pool.set(root).unwrap();
+        let root_id = pool.add(root).unwrap();
         let root = pool.get(root_id).unwrap();
         populate_tree(&mut root.write().unwrap(), 0, TREE_DEPTH, 0, &pool.clone());
 
--- a/rust/src/pool/mod.rs
+++ b/rust/src/pool/mod.rs
@@ -5,47 +5,44 @@ use std::{
 
 use crate::error::Error;
 
-pub mod node_pool_with_storage;
-
-/// An abstraction for a pool of thread-safe elements
+/// An abstraction for a pool of thread-safe items
 pub trait Pool {
-    /// The key type used to identify entries in the pool.
-    type Key;
-    /// The type of elements indexed by the pool.
+    /// The id type used to identify entries in the pool.
+    type Id;
+    /// The type of items indexed by the pool.
     type Type;
 
-    /// Retrieves an entry from the pool.
+    /// Retrieves an item from the pool.
     /// The concrete type returned by the get operation may not be the concrete type `Self::Type`
     /// but instead a wrapper type which behaves like `Self::Type`. This abstraction allows for
     /// the pool to store complex types which track the lifetime of stored objects, enabling
-    /// nested catching.
+    /// nested caching.
     fn get(
         &self,
-        id: Self::Key,
+        id: Self::Id,
     ) -> Result<PoolEntry<impl DerefMut<Target = Self::Type> + Send + Sync + 'static>, Error>;
 
-    /// Stores the value in the pool and reserves an ID for it.
-    fn set(&self, value: Self::Type) -> Result<Self::Key, Error>;
+    /// Stores the item in the pool and assigns an ID to it.
+    fn set(&self, item: Self::Type) -> Result<Self::Id, Error>;
 
-    /// Deletes the entry with the given ID from the pool
-    /// The ID may be reused in the future, when creating a new entry by calling [`Pool::set`].
-    fn delete(&self, id: Self::Key) -> Result<(), Error>;
+    /// Deletes an item with the given ID from the pool
+    /// The ID may be reused in the future, when creating a new item by calling [`Pool::set`].
+    fn delete(&self, id: Self::Id) -> Result<(), Error>;
 
     /// Flushes all pending operations to the underlying IO hierarchy.
     #[allow(dead_code)]
     fn flush(&self) -> Result<(), Error>;
 }
 
-/// An element retrieved from the pool which can be locked before reading to enable safe concurrent
-/// access. Instances of this object shall not be kept for longer than the scope where they are
-/// needed, failing to release these nodes will lead to resources leackage.
+/// An item retrieved from the pool which can be locked before reading to enable safe concurrent
+/// access.
 #[derive(Debug)]
 pub struct PoolEntry<T>(Arc<RwLock<T>>);
 
 impl<T> PoolEntry<T> {
-    /// Creates a new pool entry with the given [`NodePoolEntry`].
-    pub fn new(value: Arc<RwLock<T>>) -> Self {
-        Self(value)
+    /// Creates a new pool entry with the given [`PoolEntry`].
+    pub fn new(item: Arc<RwLock<T>>) -> Self {
+        Self(item)
     }
 
     /// Acquires a read lock on the entry.
@@ -78,22 +75,22 @@ mod tests {
     const TREE_DEPTH: u32 = 6;
     const CHILDREN_PER_NODE: u32 = 3;
 
-    /// Recursively sets up a tree with a [`FakeNodePool`] with depth `MAX_DEPTH` and `MAX_CHILDREN`
-    /// children per node. Each node's value is set to a unique number based on its position in
-    /// the tree.
+    /// Recursively sets up a tree with a [`Pool`] as backing store.
+    /// A thread is spawned for each child node to populate its subtree. Each node's value is set to
+    /// a unique number based on its position in the tree.
     fn populate_tree(
-        cur_node: &mut Node,
+        cur_node: &mut TestNode,
         depth: u32,
         max_depth: u32,
         root_id: u32,
-        pool: &Arc<impl Pool<Key = u32, Type = Node> + Send + Sync + 'static>,
+        pool: &Arc<impl Pool<Id = u32, Type = TestNode> + Send + Sync + 'static>,
     ) {
         if depth == max_depth {
             return;
         }
         let mut handles = vec![];
         for i in 1..=CHILDREN_PER_NODE {
-            let child = Node {
+            let child = TestNode {
                 value: root_id * 10 + i,
                 children: vec![],
             };
@@ -116,10 +113,10 @@ mod tests {
     }
 
     fn set_value_at_path(
-        cur_node: &Node,
+        cur_node: &TestNode,
         path: &[u32],
         value: u32,
-        pool: &Arc<impl Pool<Key = u32, Type = Node> + Send + Sync + 'static>,
+        pool: &Arc<impl Pool<Id = u32, Type = TestNode> + Send + Sync + 'static>,
     ) {
         let child_id = path
             .first()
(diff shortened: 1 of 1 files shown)


rust/src/node_pool/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl<T> NodePoolEntry<T> {
+    #[allow(dead_code)]
+    /// Creates a new pool entry with the given [`NodePoolEntry`].
+    pub fn new(value: Arc<RwLock<T>>) -> Self {
+        Self(value)
+    }
+}
+
+impl<T> Deref for NodePoolEntry<T> {
```
LuisPH3, line 41: This type does not prevent copy of the arc. Therefore, is just an alias with more steps.

What I would like to see is a type that prevents the possibility of copying the arc. If this is not possible, then we just return an alias, there is no need to dress it.
Nevertheless, I believe that it is possible having a type which implements the locking method and returns access to the internal T type.

LuisPH3 (changes requested):
Unfortunatelly this PR lacks of testing,
Such test can be the requested prof that this interface is indeed correctly locking.
Idea:
InMemoryPool, which is just a `map<NodeId, Arc<RwLock<T>>>`
A couple of tests traversing a tree and not deadlocking.

rust/src/node_pool/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn delete(&self, id: NodeId) -> Result<(), Error>;
+
+    /// Flushes all pool elements
+    fn flush(&self) -> Result<(), Error>;
+}
+
+#[allow(dead_code)]
+/// A node pool entry that can be safely shared across threads.
+#[derive(Debug)]
+pub struct NodePoolEntry<T>(Arc<RwLock<T>>);
```
psalz, line 38: The type is called the "node pool" and uses `NodeId` for getting and setting "entries" - do we really need to be generic over `T` here?

psalz (approved):
> Such test can be the requested prof that this interface is indeed correctly locking.
> Idea:
> InMemoryPool, which is just a map<NodeId, Arc<RwLock<T>>>
> A couple of tests traversing a tree and not deadlocking.

I don't really see the benefits of this: Apart from the fact that "does not deadlock" cannot be tested reliably, what would we really be testing here? Is it that `RwLock` prevents multiple threads from accessing a node concurrently? Hopefully that should be covered by the standard library. Or that using multiple nodes behind `RwLock`s concurrently does not lead to deadlocks? That entirely depends on the tree traversal algorithm. Or would we be testing that the node pool does not deadlock on concurrent get/set/delete operations? That would depend on the implementation of the pool, not the interface.

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+use crate::error::Error;
+
+/// An abstraction for a pool of thread-safe elements
+pub trait Pool {
+    /// The key type used to identify entries in the pool.
+    type Key;
+    /// The type of elements indexed by the pool.
+    type Type;
+
+    /// Retrieves an entry from the pool.
```
LuisPH3, line 15:
```suggestion
    /// Retrieves an entry from the pool.
    /// The concrete type returned by the get operation may not be the concrete type `Self::Type` but instead 
    /// a wrapper type which behaves like `Self::Type`. This abstraction allows for the pool to store complex types which track 
    /// the lifetime of stored objects, enabling nested catching. 
```

This needs some work, as the returned is the lock object, which wraps the internal type... you get the idea

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    fn get(
+        &self,
+        id: Self::Key,
+    ) -> Result<PoolEntry<impl DerefMut<Target = Self::Type> + Send + Sync + 'static>, Error>;
+
+    /// Stores the value in the pool and reserves an ID for it.
```
LuisPH3, line 21: I would say creates, at this level we do not need to know the difference between reserved and already created.

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    fn set(&self, value: Self::Type) -> Result<Self::Key, Error>;
+
+    /// Deletes the entry with the given ID from the pool
+    /// The ID may be reused in the future.
```
LuisPH3, line 25:
```suggestion
    /// The ID may be reused in the future, when creating a new entry by calling set. 
```

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    fn delete(&self, id: Self::Key) -> Result<(), Error>;
+
+    /// Flushes all pool elements
```
LuisPH3, line 28:
```suggestion
    /// signals the internal IO hierarchy to flush pending operations into persistent storage.
```

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    #[allow(dead_code)]
+    fn flush(&self) -> Result<(), Error>;
+}
+
+/// A pool entry that can be safely shared across threads.
```
LuisPH3, line 33:
```suggestion
/// An element retrieved from the pool which can be locked before reading to enable safe concurrent access.
/// Instances of this object shall not be keept for longer than the scope where they are needed, failing
/// to release these nodes will lead to resources leackage. 
```

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        for h in handles {
+            h.join().unwrap();
+        }
+    }
+
+    /// A simple in-memory pool of nodes for testing purposes.
+    struct FakeNodePool {
+        nodes: Mutex<HashMap<u32, Arc<RwLock<Node>>>>,
```
LuisPH3, line 302: nit:  you can have just a vector and use i as index

rust/src/pool/mod.rs
ihtefyw: You may reuse an old ID which technically has already been created before, depends on what we define for `create`....
Maybe `associate` is a better terminology?

rust/src/pool/mod.rs
ihtefyw: You may also have no persistent storage.
Wdyt about the current wording?

rust/src/pool/mod.rs
ihtefyw: I index from 1 in the map to make the tree generation easier. Switching to a vec would be a bit of a pain as I have to handle the indexes and I don't know how worth is that

LuisPH3 (dismissed):

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
+use std::{
+    ops::DerefMut,
+    sync::{Arc, LockResult, RwLock},
+};
+
+use crate::error::Error;
+
+/// An abstraction for a pool of thread-safe elements
```
psalz, line 8: The terminology is still inconsistent. I see "elements", "entries" and "values". The storage layer uses the term "items", maybe we should be in line with that?

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// The key type used to identify entries in the pool.
+    type Key;
+    /// The type of elements indexed by the pool.
+    type Type;
+
+    /// Retrieves an entry from the pool.
+    /// The concrete type returned by the get operation may not be the concrete type `Self::Type`
+    /// but instead a wrapper type which behaves like `Self::Type`. This abstraction allows for
+    /// the pool to store complex types which track the lifetime of stored objects, enabling
+    /// nested catching.
```
psalz, line 19:
```suggestion
    /// The concrete type returned by the get operation may not be the concrete type `Self::Type`
    /// but instead a wrapper type which behaves like `Self::Type`. This abstraction allows for
    /// the pool to store complex types which track the lifetime of stored objects, enabling
    /// nested caching.
```

Not sure what the last sentence is saying, but in general I feel like this description would better fit on the type not the method.

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+    fn get(
+        &self,
+        id: Self::Key,
+    ) -> Result<PoolEntry<impl DerefMut<Target = Self::Type> + Send + Sync + 'static>, Error>;
+
+    /// Stores the value in the pool and reserves an ID for it.
```
psalz, line 25: Inconsistent terminology: Key vs ID. Storage has a `Storage::Id`.

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+    fn set(&self, value: Self::Type) -> Result<Self::Key, Error>;
+
+    /// Deletes the entry with the given ID from the pool
+    /// The ID may be reused in the future, when creating a new entry by calling [`Pool::set`].
+    fn delete(&self, id: Self::Key) -> Result<(), Error>;
+
+    /// Flushes all pending operations to the underlying IO hierarchy.
```
psalz, line 32: What's an ID hierarchy?

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+    #[allow(dead_code)]
+    fn flush(&self) -> Result<(), Error>;
+}
+
+/// An element retrieved from the pool which can be locked before reading to enable safe concurrent
+/// access. Instances of this object shall not be kept for longer than the scope where they are
+/// needed, failing to release these nodes will lead to resources leackage.
```
psalz, line 39:
```suggestion
/// needed, failing to release these nodes will lead to resources leakage.
```

Whether or not resources will be leaked depends on the pool implementation (it might not hold on to a copy of the Arc internally), so I don't think it makes sense to prescribe the behavior on this type. I think we're being a bit too paranoid about the whole "leaking nodes" scenario.

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+        }
+        let mut handles = vec![];
+        for i in 1..=CHILDREN_PER_NODE {
+            let child = Node {
+                value: root_id * 10 + i,
+                children: vec![],
+            };
+            let child_id = pool.set(child).unwrap();
+            cur_node.children.push(child_id);
+            let child = pool.get(child_id).unwrap();
```
psalz, line 100: The `set` immediately followed by `get` is an indicator that we may want a different API here.

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+        pool::{Pool, PoolEntry},
+        storage,
+    };
+
+    const TREE_DEPTH: u32 = 6;
+    const CHILDREN_PER_NODE: u32 = 3;
+
+    /// Recursively sets up a tree with a [`FakeNodePool`] with depth `MAX_DEPTH` and `MAX_CHILDREN`
+    /// children per node. Each node's value is set to a unique number based on its position in
+    /// the tree.
```
psalz, line 81: The function operates on a pool, not a `FakeNodePool`, also the two constants no longer exist. And maybe it should mention that this is done using multiple threads - that's the main point, no?

rust/src/pool/mod.rs
```diff
@@ -0,0 +1,409 @@
...
+        }
+
+        fn flush(&self) -> Result<(), crate::error::Error> {
+            Ok(())
+        }
+    }
+
+    /// A simple tree node for testing purposes.
+    /// It stores a u32 payload a list of children node IDs.
+    struct Node {
```
psalz, line 355:
```suggestion
    struct TestNode {
```

Why not have this at the start of the module? Now I have to scroll down to figure out what this type is while reading the tests.

psalz (changes requested):

rust/src/pool/mod.rs
ihtefyw: It doesn't `insert` an ID, it just returns it. The way it is generated is implementation-defined. I think associate/reserve is generic enough.
Reserve would have been the best one but it has another meaning in the storage layer, so maybe it's better not to use it here again

rust/src/pool/mod.rs
ihtefyw: Agree on using items to align with the storage

rust/src/pool/mod.rs
ihtefyw: mhhh `assign`?

rust/src/pool/mod.rs
ihtefyw: The docstring refers to the `impl DerefMut<Target = Self::Type> ...`, not to the `PoolEntry`.
It would be nice to have an trait alias and move the comment to it. However, trait alias are experimental in stable rust

rust/src/pool/mod.rs
ihtefyw: However, IMO entry makes more clear that whatever you are using is contained within the Pool (e.g. the `PoolEntry`)
We could use `entry` for the `get` method to make explicit the ownership of the element and `item` everywhere else.
Another idea would be to rename `PoolEntry` to `PoolItem`, but IMO that's less readable.
WDYT?

rust/src/pool/mod.rs
ihtefyw: True, especially now that the lifetime of the returned type is `static` the implementer can move items out of the pool.
Removing the last sentence

rust/src/pool/mod.rs
ihtefyw: We may have a `set_and_get` or similar when we need to just want to store an intermediate result.
However, I am not sure we will need something like that in the actual VK implementation. 
I suppose we can always add it if we discover we need ti

rust/src/pool/mod.rs
psalz: Yes it doesn't insert an ID, it inserts and item into the pool and returns an ID for it.

rust/src/pool/mod.rs
```diff
@@ -1,48 +1,49 @@
...
 
-    /// Deletes the entry with the given ID from the pool
-    /// The ID may be reused in the future.
-    fn delete(&self, id: Self::Key) -> Result<(), Error>;
+    /// Deletes an item with the given ID from the pool
+    /// The ID may be reused in the future, when creating a new item by calling [`Pool::set`].
+    fn delete(&self, id: Self::Id) -> Result<(), Error>;
 
-    /// Flushes all pool elements
+    /// Flushes all pending operations to the underlying IO hierarchy.
```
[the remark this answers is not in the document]
psalz: Come to think of it, it is not even defined what an "operation" is, since the `Pool` trait does not talk about caching at all (except as an example use case). 🤷

psalz (dismissed):

LuisPH3 (dismissed):

HerbertJordan (commented):
After some offline discussion, I added a few outsider/high-level comments/ideas on how to refine the Pool trait. Please have a look, let me know what you think.

rust/src/pool/mod.rs
```diff
@@ -1,56 +1,62 @@
...
+/// IDs are not globally unique and may be reused after deletion.
+///
+/// The concrete type returned by [`Pool::get`] may not be [`Pool::Type`] but instead a wrapper type
+/// which dereferences to [`Pool::Type`]. This abstraction allows for the pool to associate metadata
+/// with each item, for example to implement smart cache eviction.
+pub trait Pool {
+    /// The id type used to identify items in the pool.
+    type Id;
+    /// The type of items indexed by the pool.
     type Type;
```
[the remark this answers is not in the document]
ihtefyw: Done

rust/src/pool/mod.rs
```diff
@@ -1,56 +1,62 @@
...
-    /// The type of elements stored in the pool. This is usually a wrapper around [`Self::Type`] and
-    /// can be dereferenced to it.
-    type StoredType;
 
-    /// Retrieves an entry from the pool.
-    fn get(&self, id: Self::Key) -> Result<PoolEntry<Self::StoredType>, Error>;
+    /// Retrieves an item from the pool, if it exists. Returns [`Error::NotFound`] otherwise.
+    fn get(
```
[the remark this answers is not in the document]
LuisPH3: If this interface constraint is imposed by the cache. Can we have other cache?

rust/src/pool/mod.rs
```diff
@@ -1,56 +1,62 @@
...
+        &self,
+        id: Self::Id,
+    ) -> Result<PoolItem<impl DerefMut<Target = Self::Type> + Send + Sync + 'static>, Error>;
 
-    /// Stores the value in the pool and reserves an ID for it.
-    fn set(&self, value: Self::Type) -> Result<Self::Key, Error>;
+    /// Stores the item in the pool and returns an ID for it.
+    fn set(&self, item: Self::Type) -> Result<Self::Id, Error>;
```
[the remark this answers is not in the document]
ihtefyw: Changed to `add`

rust/src/pool/mod.rs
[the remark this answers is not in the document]
ihtefyw: I think cloning element out of the cache is the only viable solution for a thread-safe implementation. 
Otherwise, as @LorenzSchueler stated, you would have to hold a lock on the cache to avoid data to be evicted/deleted.

PR #107, opened 2025-09-08
Address various minor issues

LorenzSchueler:
This PR refactors NodeWithMetadata and fixes a couple of typos and outdated comments in other files.

In particular, `NodeStatus` is now private and `NodeWithMetadata::new` now do longer takes a node status but just creates a node with the status clean.
The assertions with manual comparisons have been replaced by `assert_eq` and the tests that check that deref panics on deleted node use the `#[should_panic]` attribute instead of catching the panic using `catch_unwind`.


LorenzSchueler:
@psalz Done

ihtefyw (approved):

PR #114, opened 2025-09-17
Introduce cryptographic primitives for Verkle tries

psalz:
This adds three new types that form the cryptographic foundation for Verkle tries:

- `Scalar` is a value from the Banderwagon scalar field.
- `Commitment` is a vector commitment to a sequence of 256 scalar values, using the Pedersen commitment scheme.
- `Opening` is a proof demonstrating that a commitment contains certain values at specific positions.

The types are wrappers around similar types from the [`crate-crypto/rust-verkle`](https://github.com/crate-crypto/rust-verkle) repository (currently not available on crates.io), which was the best implementation of the Banderwagon curve and related functionality that I could find, even though it does not seem to be actively maintained at the moment. Notably, the `crate-crypto` organization also owns the `go-ipa` package, which is the one @HerbertJordan used for the Go reference implementation in #52.

Additional new dependencies include `ark-ff`, which is a library for finite field math operations, which `rust-verkle` also uses internally, and `const-hex`, for parsing hex-strings of expected hashes generated by the Go reference implementation.

The jury is still out on the performance of these primitives, preliminary results seem to indicate that commitments are reasonably fast (in the order of ~500us), while openings are quite slow (~30ms for a single value). However, I believe this is somewhat expected, and I don't think that we will be creating openings for individual values anyway. The library supports creating openings for multiple values and even multiple commitments at once, which should yield substantial performance gains.

NOTE that the above performance numbers are for **release builds**. In debug builds, the operations are unfortunately quite slow, leading to some tests running for 30-50s on my machine. Not sure what to do about this, other than `cargo test --release`.

Depends on #113.
Closes https://github.com/0xsoniclabs/sonic-admin/issues/267.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -29,6 +29,12 @@ zerocopy = { version = "0.8.26", features = ["derive"] }
 # error handling
 thiserror = "2.0.12"
 
+# verkle trie
+ark-ff = "0.4.2"
+banderwagon = { git = "https://github.com/crate-crypto/rust-verkle.git", rev = "e27b8b4" }
+ipa-multipoint = { git = "https://github.com/crate-crypto/rust-verkle.git", rev = "e27b8b4" }
+verkle-trie = { git = "https://github.com/crate-crypto/rust-verkle.git", rev = "e27b8b4" }
+
 [build-dependencies]
 # FFI
 bindgen = "0.72.0"
@@ -38,10 +44,13 @@ bindgen = "0.72.0"
 rstest = "0.26.1"
 rstest_reuse = "0.7.0"
 
-# mocking
+# parse hex strings into byte arrays (e.g. expected hashes)
+const-hex = "1.14.1"
+
+# automatically generate mocks from traits
 mockall = "0.13.1"
 
-# temporary files
+# create temporary files and directories for tests
 tempfile = "3.20.0"
 
 [lints.rust]
--- a/rust/src/database/verkle/crypto/commitment.rs
+++ b/rust/src/database/verkle/crypto/commitment.rs
@@ -1,0 +1,245 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+use std::sync::LazyLock;
+
+use ark_ff::{BigInteger, PrimeField};
+use banderwagon::{Element, Fr};
+use ipa_multipoint::committer::{Committer, DefaultCommitter};
+use verkle_trie::constants::CRS;
+use zerocopy::{FromBytes, Immutable, IntoBytes, Unaligned};
+
+use crate::database::verkle::crypto::Scalar;
+
+/// A vector commitment to a sequence of 256 scalar values, using the Pedersen commitment scheme
+/// on the Banderwagon curve.
+///
+/// Commitments can be de-/serialized from/to bytes using zerocopy.
+/// Note that not all byte arrays correspond to valid commitments, i.e., points in the Banderwagon
+/// prime subgroup. Performing any operations on invalid commitments will produce indeterminate
+/// results (garbage in, garbage out).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, FromBytes, IntoBytes, Immutable, Unaligned)]
+#[repr(C)]
+pub struct Commitment {
+    // We store a byte representation to allow for easy serialization using zerocopy.
+    // Note that conversion from/to banderwagon::Element is NOT trivial (due to compression).
+    // TODO: Benchmark performance impact (https://github.com/0xsoniclabs/sonic-admin/issues/373).
+    point_bytes: [u8; 32],
+}
+
+// Creating the committer is very expensive (in the order of seconds!), so we cache it.
+static COMMITTER: LazyLock<DefaultCommitter> = LazyLock::new(|| DefaultCommitter::new(&CRS.G));
+
+impl Commitment {
+    /// Creates a commitment to the given sequence of scalar values.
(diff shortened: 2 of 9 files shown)


rust/src/database/verkle/crypto/opening.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Creates a new opening for a given commitment, proving that it contains the
+    /// value at the specified position. The opening can then be verified using
+    /// [`Self::verify_single`].
+    ///
+    /// For the opening to be valid, the `values` slice must contain the same sequence
+    /// of scalars that were used to create the commitment.
+    ///
+    /// NOTE: This is fairly expensive operation (tens of milliseconds).
+    /// Prefer to open multiple values or even multiple commitments at once, if possible.
+    pub fn open_single(commitment: &Commitment, values: &[Scalar], position: u8) -> Self {
```
LorenzSchueler, line 40: nit: `Opening::open_single` sounds like you open the opening, but you are actually creating it. Maybe `for_single` would be more fitting. Same for `open_all`.

rust/src/database/verkle/crypto/opening.rs
```diff
@@ -0,0 +1,186 @@
...
+
+            assert!(
+                !opening.verify_single(&commitment, i as u8, &old_value),
+                "verified outdated value at position {i}"
+            );
+        }
+    }
+
+    #[test]
+    fn open_all_can_be_used_to_proof_all_values() {
```
LorenzSchueler, line 170: Should we maybe prefix all slow tests with `slow_*`? During development if you work on something different you can then run ` cargo test -- --skip slow_`.

rust/src/database/verkle/crypto/commitment.rs
```diff
@@ -0,0 +1,241 @@
...
+            .to_bytes_le()
+            .try_into()
+            .expect("scalar field element should be 32 bytes")
+    }
+
+    /// Returns a compressed 32-byte representation of the commitment.
+    /// Used as the state root commitment in Verkle tries.
+    pub fn compress(&self) -> [u8; 32] {
+        self.point_bytes
+    }
```
LorenzSchueler, line 80: @HerbertJordan: Hash and compress produce different values and are used differently, but do you know why we could not just use one of these everywhere?

rust/Cargo.toml
```diff
@@ -29,15 +29,19 @@ zerocopy = { version = "0.8.26", features = ["derive"] }
...
 # FFI
 bindgen = "0.72.0"
 
 [dev-dependencies]
-# mocking
+const-hex = "1.14.1"
 mockall = "0.13.1"
-
-# temporary files
 tempfile = "3.20.0"
```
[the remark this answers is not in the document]
psalz: I honestly don't see much value in having categories like "temporary files" above a crate called "tempfile", in particular if each category only contains a single crate, which was the case so far. I've brought them back, but in a more descriptive form. Maybe we should even have one comment per crate, instead of thinking of them as groups. WDYT?

rust/src/database/verkle/crypto/scalar.rs
```diff
@@ -0,0 +1,163 @@
...
+        Scalar(Fr::from(0))
+    }
+
+    /// Creates a new scalar from the given little-endian byte slice.
+    /// At most 31 bytes are read from the slice, shorter slices are right-padded with zeros.
+    pub fn from_le_bytes(bytes: &[u8]) -> Self {
+        let mut arr = [0u8; 32];
+        let len = bytes.len().min(31);
+        arr[..len].copy_from_slice(&bytes[..len]);
+        Scalar(Fr::from_le_bytes_mod_order(&arr))
```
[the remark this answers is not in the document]
psalz: I am not sure what the ramifications of this would be, but it's again also just a hypothetical because we don't use more than 16 bytes. In any case, the copy was unnecessary and I'm now simply forwarding `bytes[...bytes.len().min(31)]`.

rust/src/database/verkle/crypto/scalar.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Sets the 128th bit of the scalar to 1.
+    /// This is a convenience method used for computing the commitment of a
+    /// leaf node in Ethereum Verkle tries, where the 128th bit indicates
+    /// that a value has been set before (to distinguish it from the default
+    /// value, which is needed for future state expiry schemes).
+    pub fn set_bit128(&mut self) {
+        let mut bytes = self.0.into_bigint().to_bytes_be();
+        bytes[15] |= 0x01;
+        self.0 = Fr::from_be_bytes_mod_order(&bytes);
```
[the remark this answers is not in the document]
psalz: Yes good point, changed.

rust/src/database/verkle/crypto/opening.rs
psalz: Yes I thought `new_all` sounds weird, but `for_all` is not too bad - renamed!

rust/src/database/verkle/crypto/opening.rs
```diff
@@ -0,0 +1,186 @@
...
+/// A proof demonstrating that a [`Commitment`] contains certain values at specific positions.
+///
+/// It uses the Inner Product Argument (IPA) proof system for Pedersen commitments.
+/// Details: https://dankradfeist.de/ethereum/2021/07/27/inner-product-arguments.html
+pub struct Opening {
+    pub(crate) proof: ipa_multipoint::multiproof::MultiPointProof,
+}
+
+#[cfg_attr(not(test), expect(unused))]
+impl Opening {
```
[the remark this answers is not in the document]
psalz: Nothing, they just don't verify. I've added two test cases that demonstrate this. We can consider having different types for these, but the whole proof format is still in flux anyway, until we figure out how Ethereum does witness proofs.

rust/src/database/verkle/crypto/opening.rs
LorenzSchueler: also fine

LorenzSchueler:
I looked at the diff, which looks good. I still have to go through the changes in total once again, though.
One thing I noticed is that there are still a bunch of tests that need ~20s in debug mode. Please put those also in the `slow` module.

PR #116, opened 2025-09-18
Fix code splitting edge case in Go Verkle implementation

psalz:
This fixes an edge case in the Go implementation of code splitting for Verkle tries, where codes that are shorter than expected (based on PUSH instructions towards the end) would result in chunks with too many bytes marked as data (32, which is not even  possible).

I've added two new test cases for code splitting, and adjusted one of the cases in the state root comparison with Geth to make sure they behave the same way.

### DIFF
--- a/go/database/vt/memory/code.go
+++ b/go/database/vt/memory/code.go
@@ -40,7 +40,7 @@ func splitCode(code []byte) []chunk {
 	chunks := make([]chunk, 0, len(code)/32+1)
 	for i := 0; len(code) > 0; i++ {
 		next := chunk{}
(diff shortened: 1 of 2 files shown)


LorenzSchueler (approved):
Good catch!

kjezek (approved):

psalz:
We found another bug in code splitting during the review for #119, where codes that end in a PUSH32 instruction followed by the correct number of data bytes (i.e., 32), could result in a chunk that has 32 bytes marked as data (where 31 is the maximum). This is fixed by limiting the loop over `j` to at most 31 iterations.

LorenzSchueler (approved):

PR #117, opened 2025-09-19
Use rstest for file backend tests

LorenzSchueler:
This PR uses [rstest](https://crates.io/crates/rstest) and [rstest_reuse](https://crates.io/crates/rstest_reuse) for the file backend tests.

`rstest` allows creating parameterized test, so that instead of having a test function which loops over all cases, the macro generates one test for each case. This makes debugging easier, because if a test fails, the test name already includes the case which failed.

`restes_reuse` can be used if multiple test operate on the same cases. By using it, the cases have to be written only once and can be reused for many tests.

Because we use very long and descriptive test names, I think it would be useful to allow non-snake-case names for easier reading. We should still use consistent formatting, though. I would suggest staying with all lower case but allowing double `__`.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -56,6 +56,8 @@ dependencies = [
  "bindgen",
  "dashmap",
  "mockall",
+ "rstest",
+ "rstest_reuse",
  "tempfile",
  "thiserror",
  "zerocopy",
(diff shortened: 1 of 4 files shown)


psalz (dismissed):
I'd like to see an example of the double underscore being used in a test name, but it doesn't seem like there is one..?

LorenzSchueler:
the file_backend ones now have one e.g. `storage::file::file_backend::tests::access_same_page_in_parallel_does_not_deadlock::case_4_page_cached_file__no_seek_file`

PR #120, opened 2025-09-19
Add simple in-memory Verkle trie implementation

psalz:
This adds a simple in-memory implementation of the *Verkle trie*, [proposed by Ethereum](https://blog.ethereum.org/2021/12/02/verkle-tree-structure) as a potential replacement for the Merkle-Patricia Trie (MPT).

The Verkle trie has several advantages compared to the MPT:
- Simpler structure: Only two types of nodes (inner and leaf)
- No RLP encoding
- Single unified tree for accounts, storage and contract codes
- 256 children per node, leading to a much shallower tree
- Smaller proofs based on vector commitments

The major potential downside of Verkle tries is that the underlying elliptic curve cryptographic operations are substantially more expensive than e.g. MPT's Keccak-256. We'll have to properly quantify this using extensive benchmarking.

The implementation provided in this PR is a very simple in-memory tree, i.e., it does not store any data on disk. It is not  meant for production use, but rather as a reference, and for validation, prototyping and testing.

The implementation has several shortcomings:
  - Total tree size is limited by available memory.
  - No concurrency support (single lock on root node).
  - Not optimized for memory usage (all nodes store 256 children / values).

Depends on #119.

### DIFF
--- a/rust/src/database/verkle/compute_commitment.rs
+++ b/rust/src/database/verkle/compute_commitment.rs
@@ -1,0 +1,50 @@
+use crate::{
+    database::verkle::crypto::{Commitment, Scalar},
+    types::Value,
+};
+
+/// Computes the commitment of a leaf node.
+///
+/// Since [`crate::database::verkle::crypto::Scalar`] cannot safely represent 32 bytes,
+/// the 256 32-bit values are split into two interleaved sets of 16 byte values, on which
+/// commitments C1 and C2 are computed separately:
+///
+///   C1 = Commit([  v[0][..16]),   v[0][16..]),   v[1][..16]),   v[1][16..]), ...])
+///   C2 = Commit([v[128][..16]), v[128][16..]), v[129][..16]), v[129][16..]), ...])
+///
+/// The final commitment of a leaf node is then computed as follows:
+///
+///    C = Commit([1, stem, C1, C2])
+///
+/// For details on the commitment procedure, see
+/// https://blog.ethereum.org/2021/12/02/verkle-tree-structure#commitment-to-the-values-leaf-nodes
+#[cfg_attr(not(test), expect(unused))]
+pub fn compute_leaf_node_commitment(
+    input_values: &[Value; 256],
+    used_bits: &[u8; 256 / 8],
+    stem: &[u8; 31],
+) -> Commitment {
+    let mut values = vec![vec![Commitment::default().to_scalar(); 256]; 2];
+    for (i, value) in input_values.iter().enumerate() {
+        let mut lower = Scalar::from_le_bytes(&value[..16]);
+        let upper = Scalar::from_le_bytes(&value[16..]);
+
+        if used_bits[i / 8] & (1 << (i % 8)) != 0 {
+            lower.set_bit128();
+        }
+
+        values[i / 128][(2 * i) % 256] = lower;
+        values[i / 128][(2 * i + 1) % 256] = upper;
+    }
+
+    let c1 = Commitment::new(&values[0]);
+    let c2 = Commitment::new(&values[1]);
+
+    let combined = vec![
+        Scalar::from(1),
+        Scalar::from_le_bytes(stem),
+        c1.to_scalar(),
+        c2.to_scalar(),
+    ];
+    Commitment::new(&combined)
+}
--- a/rust/src/database/verkle/variants/simple/node.rs
+++ b/rust/src/database/verkle/variants/simple/node.rs
@@ -1,0 +1,571 @@
+#[cfg(test)]
+use crate::{database::verkle::compute_commitment::compute_leaf_node_commitment, types::Key};
+use crate::{database::verkle::crypto::Commitment, types::Value};
+
+/// A node in the simple in-memory Verkle trie.
+#[cfg_attr(not(test), expect(unused))]
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug)]
+pub enum Node {
+    Empty,
+    Inner(InnerNode),
+    Leaf(LeafNode),
+}
+
+#[cfg(test)]
+impl Node {
+    /// Returns the value associated with the given key, or the default value if
+    /// the key does not exist.
+    pub fn get(&self, key: &Key, depth: u8) -> Value {
+        match self {
+            Node::Empty => Value::default(),
+            Node::Inner(inner) => inner.get(key, depth),
+            Node::Leaf(leaf) => leaf.get(key),
+        }
+    }
+
+    /// Sets the value for the given key.
+    /// Consumes the node and returns an updated version.
+    pub fn set(self, key: &Key, depth: u8, value: &Value) -> Node {
+        match self {
+            Node::Empty => {
+                // While conceptually it would suffice to create a leaf node here,
+                // Geth always creates an inner node (and we want to stay compatible).
+                let inner = InnerNode::new();
+                inner.set(key, depth, value)
+            }
+            Node::Inner(inner) => inner.set(key, depth, value),
+            Node::Leaf(leaf) => leaf.set(key, depth, value),
+        }
+    }
+
+    /// Computes and returns the commitment of this node.
+    ///
+    /// If the commitment is already up to date, it is returned without recomputation.
+    pub fn commit(&mut self) -> Commitment {
+        match self {
+            Node::Empty => Commitment::default(),
+            Node::Inner(inner) => inner.commit(),
+            Node::Leaf(leaf) => leaf.commit(),
+        }
+    }
+
+    /// Returns the current commitment of this node without recomputing it.
+    fn get_commitment(&self) -> Commitment {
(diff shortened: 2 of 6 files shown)


rust/src/database/verkle/compute_commitment.rs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        values[i / 128][(2 * i) % 256] = lower;
+        values[i / 128][(2 * i + 1) % 256] = upper;
+    }
+
+    let c1 = Commitment::new(&values[0]);
+    let c2 = Commitment::new(&values[1]);
+
+    let combined = vec![
```
LorenzSchueler, line 53:
```suggestion
    let combined = [
```

rust/src/database/verkle/compute_commitment.rs
```diff
@@ -0,0 +1,60 @@
...
+///   C2 = Commit([v[128][..16]), v[128][16..]), v[129][..16]), v[129][16..]), ...])
+///
+/// The final commitment of a leaf node is then computed as follows:
+///
+///    C = Commit([1, stem, C1, C2])
+///
+/// For details on the commitment procedure, see
+/// https://blog.ethereum.org/2021/12/02/verkle-tree-structure#commitment-to-the-values-leaf-nodes
+#[cfg_attr(not(test), expect(unused))]
+pub fn compute_leaf_node_commitment(
```
LorenzSchueler, line 32: This function needs tests.

ihtefyw (commented):
Looks good. I agree with @LorenzSchueler taht the `compute_leaf_commitment` function needs testing

rust/src/database/verkle/compute_commitment.rs
psalz: I added a test comparing against a hash generated with the Go reference implementation.

rust/src/database/verkle/variants/simple/node.rs
```diff
@@ -0,0 +1,581 @@
...
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+#[cfg(test)]
+use crate::{database::verkle::compute_commitment::compute_leaf_node_commitment, types::Key};
+use crate::{database::verkle::crypto::Commitment, types::Value};
+
+/// A node in the simple in-memory Verkle trie.
+#[cfg_attr(not(test), expect(unused))]
+#[allow(clippy::large_enum_variant)]
```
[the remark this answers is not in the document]
LorenzSchueler: With an array they are also co-located contiguously in memory 🤔

rust/src/database/verkle/compute_commitment.rs
```diff
@@ -50,11 +50,47 @@ pub fn compute_leaf_node_commitment(
...
+    use crate::database::verkle::test_utils::FromIndexValues;
+
+    #[test]
+    fn compute_leaf_node_commitment_produces_expected_values() {
+        {
+            let values = [Value::default(); 256];
+            let used_bits = [0; 256 / 8];
+            let stem = [0u8; 31];
+            let commitment = compute_leaf_node_commitment(&values, &used_bits, &stem);
+            let expected = Commitment::new(&[Scalar::from(1)]);
```
LorenzSchueler, line 74: I'm probably overlooking something, but why is the commitment of empty values the same as the commitment of 1?

rust/src/database/verkle/compute_commitment.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// For details on the commitment procedure, see
+/// https://blog.ethereum.org/2021/12/02/verkle-tree-structure#commitment-to-the-values-leaf-nodes
+#[cfg_attr(not(test), expect(unused))]
+pub fn compute_leaf_node_commitment(
+    input_values: &[Value; 256],
+    used_bits: &[u8; 256 / 8],
+    stem: &[u8; 31],
+) -> Commitment {
+    let mut values = vec![vec![Commitment::default().to_scalar(); 256]; 2];
```
LorenzSchueler, line 37: nit:

```suggestion
    let mut values = [[Commitment::default().to_scalar(); 256]; 2];
```

rust/src/database/verkle/variants/simple/node.rs
```diff
@@ -0,0 +1,581 @@
...
+
+    /// Sets the value for the given key.
+    /// Consumes the node and returns an updated version.
+    pub fn set(self, key: &Key, depth: u8, value: &Value) -> Node {
+        match self {
+            Node::Empty => {
+                // While conceptually it would suffice to create a leaf node here,
+                // Geth always creates an inner node (and we want to stay compatible).
+                let inner = InnerNode::new();
+                inner.set(key, depth, value)
```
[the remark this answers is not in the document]
LorenzSchueler: See suggestion below.

LorenzSchueler, line 45:
```suggestion
                if depth == 0 {
                    // While conceptually it would suffice to create a leaf node here,
                    // Geth always creates an inner node (and we want to stay compatible).
                    let inner = InnerNode::new();
                    inner.set(key, depth, value)
                } else {
                    let leaf = LeafNode::new(key);
                    leaf.set(key, depth, value)
                }
```

rust/src/database/verkle/variants/simple/node.rs
```diff
@@ -0,0 +1,581 @@
...
+    /// If no child exists at that position, a new leaf node is created.
+    ///
+    /// Consumes the node and returns an updated version.
+    pub fn set(mut self, key: &Key, depth: u8, value: &Value) -> Node {
+        self.commitment_dirty = true;
+
+        let pos = key[depth as usize];
+        if matches!(self.children[pos as usize], Node::Empty) {
+            self.children[pos as usize] = Node::Leaf(LeafNode::new(key));
+        }
```
LorenzSchueler, line 131:
```suggestion
```
see comment above

LorenzSchueler (changes requested):

rust/src/database/verkle/variants/simple/node.rs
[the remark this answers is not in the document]
psalz: Maybe I'm missing something, but with a vector of values we have one pointer indirection to a contiguous allocation of data `ptr -> [node, node, node, ...]`, whereas with a array of boxes we have a contiguous allocation of pointers to nodes `[ptr -> node, ptr -> node, ptr -> node, ...]`, no? Depending on the allocator the nodes in the latter might still end up being contiguous in memory, but it's not guaranteed.

rust/src/database/verkle/compute_commitment.rs
psalz: It just works out that way because C1, C2 and stem all end up being the zero commitment. The final commitment is then `commitment(1, stem, C1, C2)`.

LorenzSchueler: Ah that magic `1` 😆

rust/src/database/verkle/variants/simple/node.rs
psalz: I like it!

rust/src/database/verkle/variants/simple/node.rs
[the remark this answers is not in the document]
LorenzSchueler: Ah ok, you meant the actual data of the children.
I don't know if we really have an access pattern that traverses through the children of a node because with vector commitments we do **not** need the data of other children when updating the commitment of the parent when one child changed.
But if you want to keep it this way for now, that's fine with me.

rust/src/database/verkle/variants/simple/node.rs
```diff
@@ -0,0 +1,583 @@
...
+        let mut children = Vec::with_capacity(256);
+        for _ in 0..256 {
+            children.push(Node::Empty);
+        }
+        InnerNode {
+            children,
+            commitment: Commitment::default(),
+            commitment_dirty: true,
+        }
+    }
```
LorenzSchueler, line 109: Sorry for my urge of having to encode everything in the type system, but if we keep the allocation here then please use a boxed array instead.
```suggestion
pub struct InnerNode {
    children: Box<[Node; 256]>,
    commitment: Commitment,
    commitment_dirty: bool,
}

#[cfg(test)]
impl InnerNode {
    /// Creates a new inner node without any children.
    fn new() -> Self {
        let children = Box::new(std::array::from_fn(|_| Node::Empty));
        InnerNode {
            children,
            commitment: Commitment::default(),
            commitment_dirty: true,
        }
    }
```

LorenzSchueler (approved):

rust/src/database/verkle/variants/simple/node.rs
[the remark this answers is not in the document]
psalz: Yes I also don't expect it to make a difference, but this way we can check!

PR #125, opened 2025-09-23
Make file backend benchmarks parallel

LorenzSchueler:
This PR improves the file backend benchmarks by calling the file backend from multiple threads in parallel. The iterations are distributed round-robin. Because criterion has no built-in support for this use case, the `iter_custom` method is used, and the time measurements are implemented by hand. To avoid measuring the spawning of the threads, all threads busy wait on an atomic, which is flipped once all threads have been spawned and the time measurement started.

Additionally, it is now ensured, that the iteration counter is **not** reset after each run invocation of `iter_custom` (criterion uses many samples for both the warmup and benchmarking phase) but instead each run starts where the last one finished off. This is important to make sure that the first invocation does not load everything into the cache and the following ones have all data cached already.

Part of https://github.com/0xsoniclabs/sonic-admin/issues/367

### DIFF
--- a/rust/benches/file_backend.rs
+++ b/rust/benches/file_backend.rs
@@ -255,6 +255,7 @@ fn file_backend_benchmark(
                             while !start_toggle.load(Ordering::Acquire) {
                                 std::hint::spin_loop();
                             }
+                            let start = Instant::now();
                             for iter in ((completed_iterations + thread as u64)
                                 ..(completed_iterations + iterations))
                                 .step_by(threads)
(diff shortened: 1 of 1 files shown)


rust/benches/file_backend.rs
```diff
@@ -201,44 +206,74 @@ fn file_backend_benchmark_matrix(c: &mut Criterion) {
...
+                                operation.execute(backend.deref(), &mut data, offset, iter);
+                            }
+                        }));
+                    }
+                    let start = Instant::now();
+                    start_toggle.store(true, Ordering::Release);
+
+                    for handle in handles {
+                        handle.join().unwrap();
+                    }
```
psalz, line 272: Not sure if it matters, but instead of measuring the join time here, each thread could measure it's own time and then do a `AtomicU64::fetch_max`.

psalz (approved):

rust/benches/file_backend.rs
LorenzSchueler: Good point. I opted for last_end - first_start.

PR #127, opened 2025-09-24
Improve ReuseListFile

LorenzSchueler:
This PR adds the `ReuseListFile` which will replace the `CachedReuseListFile` (in `node_file_storage.rs`) in the next PR.
`ReuseListFile` serves the same purpose, but unlike `CachedReuseListFile` which is only a struct without any logic, it implements all necessary operation so that the logic can be pulled out of `NodeFileStorage`. It also implements freezing part of the list which is later needed for snapshots.


rust/src/storage/file/reuse_list_file.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    #[test]
+    fn write_fails_if_file_cannot_be_written() {
+        let dir = TestDir::try_new(Permissions::ReadWrite).unwrap();
+        let path = dir.path().join("reuse_list");
+
+        File::create(path.as_path()).unwrap();
+
+        let cached_file = ReuseListFile {
+            file: File::open(path.as_path()).unwrap(),
```
psalz, line 177:
```suggestion
            file: File::open(path.as_path()).unwrap(), // read-only
```

rust/src/storage/file/reuse_list_file.rs
```diff
@@ -0,0 +1,251 @@
...
+        self.cache.pop()
+    }
+
+    /// Pushes an index to the cache.
+    pub fn push(&mut self, idx: u64) {
+        self.cache.push(idx);
+    }
+
+    /// Returns a slice of the cached indices, including the frozen part.
+    pub fn as_slice(&self) -> &[u64] {
```
psalz, line 79: What is the use case for this?

PR #130, opened 2025-09-26
Add Carmen state implementation based on in-memory Verkle trie

psalz:
This adds an implementation of the `CarmenState` trait based on the simple in-memory Verkle trie introduced with #120. The implementation of `VerkleTrieState` is generic over the trie implementation, and the tests are already set up in a way that supports additional implementations in the future.

The state interface is hooked up all the way to the FFI boundary, meaning that it can theoretically already be called from Go. We are however still missing some changes to the tests on the Go side to properly support S6 (cf. #80). The way the state is exposed through FFI is still a bit rudimentary: The interface expects a `Box<dyn CarmenState>`, which conceptually should be some kind of view into the database at the current block height. Since no such mechanism exists yet, the simplest solution was to have `CarmenS6Db` contain an `Arc<CarmenState>` which itself also implements the `CarmenState` trait and is then cloned into the returned box.

This introduces [sha3](https://crates.io/crates/sha3) as a new dependency, which we need to compute the `Keccak-256` hash of contract codes.

Depends on #120.

### DIFF
--- a/rust/src/database/verkle/state.rs
+++ b/rust/src/database/verkle/state.rs
@@ -31,18 +31,18 @@ const EMPTY_CODE_HASH: Hash = [
 ];
 
 /// An implementation of [`CarmenState`] that uses a Verkle trie as the underlying data structure.
-pub struct VerkleTrieState<T: VerkleTrie> {
+pub struct VerkleTrieCarmenState<T: VerkleTrie> {
     trie: T,
 }
 
-impl VerkleTrieState<SimpleInMemoryVerkleTrie> {
+impl VerkleTrieCarmenState<SimpleInMemoryVerkleTrie> {
     pub fn new() -> Self {
         let trie = SimpleInMemoryVerkleTrie::new();
         Self { trie }
     }
 }
 
-impl<T: VerkleTrie> CarmenState for VerkleTrieState<T> {
+impl<T: VerkleTrie> CarmenState for VerkleTrieCarmenState<T> {
     fn account_exists(&self, addr: &Address) -> Result<bool, Error> {
         Ok(self.get_code_hash(addr)? != Hash::default())
     }
@@ -175,7 +175,7 @@ mod tests {
 
     #[rstest_reuse::template]
     #[rstest::rstest]
-    #[case::simple_in_memory(Box::new(VerkleTrieState::<SimpleInMemoryVerkleTrie>::new()) as Box<dyn CarmenState>)]
+    #[case::simple_in_memory(Box::new(VerkleTrieCarmenState::<SimpleInMemoryVerkleTrie>::new()) as Box<dyn CarmenState>)]
     fn all_state_impls(#[case] state: Box<dyn CarmenState>) {}
 
     #[test]
@@ -187,7 +187,7 @@ mod tests {
 
     #[test]
     fn new_creates_empty_state() {
-        let state = VerkleTrieState::<SimpleInMemoryVerkleTrie>::new();
+        let state = VerkleTrieCarmenState::<SimpleInMemoryVerkleTrie>::new();
         assert_eq!(state.get_hash().unwrap(), Hash::default());
     }
 
--- a/rust/src/database/mod.rs
+++ b/rust/src/database/mod.rs
@@ -10,4 +10,4 @@
 
 mod verkle;
 
-pub use verkle::{SimpleInMemoryVerkleTrie, VerkleTrieState};
+pub use verkle::{SimpleInMemoryVerkleTrie, VerkleTrieCarmenState};
--- a/rust/src/database/verkle/mod.rs
+++ b/rust/src/database/verkle/mod.rs
@@ -17,5 +17,5 @@ mod test_utils;
 mod variants;
 mod verkle_trie;
 
-pub use state::VerkleTrieState;
+pub use state::VerkleTrieCarmenState;
 pub use variants::SimpleInMemoryVerkleTrie;
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -11,7 +11,7 @@
 
 use std::{mem::MaybeUninit, ops::Deref, sync::Arc};
 
-use crate::{database::VerkleTrieState, error::Error, types::*};
+use crate::{database::VerkleTrieCarmenState, error::Error, types::*};
 
 mod database;
 mod error;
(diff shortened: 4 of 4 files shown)


rust/src/ffi/exported.rs
```diff
@@ -1168,8 +1168,8 @@ mod tests {
 
     #[test]
     fn carmen_rust_open_database_returns_non_null_pointers() {
-        let live_impls = [LiveImpl::Memory, LiveImpl::File, LiveImpl::LevelDb];
-        let archive_impls = [ArchiveImpl::None, ArchiveImpl::LevelDb, ArchiveImpl::Sqlite];
+        let live_impls = [LiveImpl::Memory];
+        let archive_impls = [ArchiveImpl::None];
```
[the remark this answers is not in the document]
psalz: As discussed in person, I've brought back the different variants but now test whether the call to `Carmen_Rust_OpenDatabase` either returns a valid database pointer or an `Result_kResult_UnsupportedImplementation`. I've introduced a new corresponding enum variant for `Error`.

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,527 @@
...
+        },
+        variants::SimpleInMemoryVerkleTrie,
+        verkle_trie::VerkleTrie,
+    },
+    error::Error,
+    types::{Address, Hash, Key, Nonce, U256, Update, Value},
+};
+
+/// An implementation of [`CarmenState`] that uses a Verkle trie as the underlying data structure.
+pub struct VerkleTrieState<T: VerkleTrie + Send + Sync> {
```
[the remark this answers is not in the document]
psalz: Seems to me like Clippy should catch that...

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,528 @@
...
+        },
+        variants::SimpleInMemoryVerkleTrie,
+        verkle_trie::VerkleTrie,
+    },
+    error::Error,
+    types::{Address, Hash, Key, Nonce, U256, Update, Value},
+};
+
+/// An implementation of [`CarmenState`] that uses a Verkle trie as the underlying data structure.
+pub struct VerkleTrieState<T: VerkleTrie> {
```
LorenzSchueler, line 29: I find the naming a little bit confusing. Just given then names `VerkleTrie` and `VerkleTrieState` it is not obvious to me what the difference is supposed to be.
Should we rename `VerkleTrieState` maybe to something like `VerkleTrieWithEmbedding` or `EmbeddedVerkleTrie`? This is essentially what this type does...

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,528 @@
...
+                .apply_block_update(
+                    0,
+                    Update {
+                        slots: &[SlotUpdate { addr, key, value }],
+                        ..Default::default()
+                    },
+                )
+                .unwrap();
+        };
+
```
LorenzSchueler, line 407: set_balance and set_nonce are freestanding functions, so let's keep it consistent.
```suggestion
    fn set_storage(state: &dyn CarmenState, addr: Address, key: Key, value: Value) {
        state
            .apply_block_update(
                0,
                Update {
                    slots: &[SlotUpdate { addr, key, value }],
                    ..Default::default()
                },
            )
            .unwrap();
    }

    #[rstest_reuse::apply(all_state_impls)]
    fn can_store_and_retrieve_storage_values(#[case] state: Box<dyn CarmenState>) {
```

LorenzSchueler (dismissed):

rust/src/database/verkle/state.rs
ihtefyw: What about `VerkleTrieCarmenState`? And in the future we would have `BinaryTreeCarmenState`, `MPTCarmenState`, etc.?

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,528 @@
...
+        state
+            .apply_block_update(
+                0,
+                Update {
+                    nonces: &[NonceUpdate { addr, nonce }],
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+    }
```
ihtefyw, line 192: Please move it to the bottom of the test module, we usually put utility functions at the end

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,528 @@
...
+        state
+            .apply_block_update(
+                0,
+                Update {
+                    balances: &[BalanceUpdate { addr, balance }],
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+    }
```
ihtefyw, line 235: Same, move it down

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,528 @@
...
+        state
+            .apply_block_update(
+                0,
+                Update {
+                    codes: vec![CodeUpdate { addr, code }],
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+    }
```
ihtefyw, line 303: Same, move it down

ihtefyw (changes requested):

psalz:
I've amended the implementation to now also support account creation and `account_exists`, since this is required to pass various tests on the Go side until we move that functionality further up the stack (cf. https://github.com/0xsoniclabs/sonic-admin/issues/383).

rust/src/database/verkle/state.rs
```diff
@@ -0,0 +1,589 @@
...
+impl VerkleTrieState<SimpleInMemoryVerkleTrie> {
+    pub fn new() -> Self {
+        let trie = SimpleInMemoryVerkleTrie::new();
+        Self { trie }
+    }
+}
+
+impl<T: VerkleTrie> CarmenState for VerkleTrieState<T> {
+    fn account_exists(&self, addr: &Address) -> Result<bool, Error> {
+        Ok(self.get_code_hash(addr)? != Hash::default())
```
LorenzSchueler, line 47: I have not looked into the exact semantics, but why not check the nonce instead? EOAs may not have code.

rust/src/database/verkle/state.rs
LorenzSchueler: I'd be also fine with that.

LorenzSchueler (changes requested):

PR #131, opened 2025-09-29
Decouple Storage Layers

LorenzSchueler:
This PR decouples the storage layers, in places where they were still tightly coupled.
Most notably, `FileStorageManager` is now generic over 3 underlying storages. This has several advantages:
- in the tests, the underlying storages can now use mocks
  - with mocks we can now actually test the main functionality of the manager, which is to forward calls to the corresponding underlying storage, depending on the node type. Before, this was only indirectly possible by observing the side effects of calling the `NodeFileStorage`.
- it is possible to use different storages, depending on the node type (e.g. `NodeFileStorage<FullInnerNode, NoSeekFile>` and `NodeFileStorage<SparseInnerNode<2>, PageCachedFile<NoSeekFile>>`)

Also, the mocks are now generated using `mockall::mock` instead of `mockall::automock`. This has 2 reasons. By using `mock` the definition of the mock is in the same file in which it is used and can have a more suitable name. The second one is that in an upcoming PR which implements checkpoints, the mocks will have to implement multiple traits, which is not possible with `automock`.


rust/src/storage/file/file_storage_manager.rs
```diff
@@ -122,20 +136,24 @@ where
...
-    type FileStorageManager = super::FileStorageManager<SeekFile>;
-    type NodeFileStorage = super::NodeFileStorage<InnerNode, SeekFile>;
+    use crate::{
+        storage::file::{NodeFileStorage, SeekFile},
+        types::NodeId,
+    };
 
     #[test]
-    fn open_creates_files_if_they_do_not_exist() {
+    fn open_creates_directory_and_opens_committed_checkpoint_and_calls_open_on_all_storages() {
```
ihtefyw, line 150:
```suggestion
    fn open_creates_directory_and_calls_open_on_all_storages() {
```
There are still no checkpoints

rust/src/storage/file/file_storage_manager.rs
```diff
@@ -185,210 +203,241 @@ mod tests {
...
+            let leaf_node_2 = SparseLeafNode::default();
             storage
-                .get(NodeId::from_idx_and_node_type(0, NodeType::Leaf2))
-                .unwrap(),
-            Node::Leaf2(Box::new(leaf_node_2))
-        );
-        assert_eq!(
+                .leaf_nodes_2
+                .expect_get()
+                .withf(move |id| *id == leaf_node_2_id.to_index())
```
ihtefyw, line 266:
```suggestion
                .with(eq(leaf_node_2_id.to_index()))
```

rust/src/storage/file/file_storage_manager.rs
```diff
@@ -185,210 +203,241 @@ mod tests {
...
+            let inner_node = InnerNode::default();
             storage
-                .get(NodeId::from_idx_and_node_type(0, NodeType::Inner))
-                .unwrap(),
-            Node::Inner(Box::new(inner_node))
-        );
-        assert_eq!(
+                .inner_nodes
+                .expect_get()
+                .withf(move |id| *id == inner_node_id.to_index())
```
ihtefyw, line 248:
```suggestion
                .with(eq(inner_node_id.to_index()))
```

ihtefyw (commented):
Looks good, some minor issues to address

rust/src/storage/file/file_storage_manager.rs
```diff
@@ -400,99 +433,89 @@ mod tests {
...
+                .returning(move |_| Ok(()));
+            assert!(storage.delete(leaf_node_256_id).is_ok());
+        }
     }
 
     #[test]
-    fn flush_flushes_all_node_file_storages() {
-        let dir = tempfile::tempdir().unwrap();
-        let path = dir.path();
+    fn flush_calls_flushes_on_all_storages() {
```
psalz, line 485:
```suggestion
    fn flush_calls_flush_on_all_storages() {
```

rust/src/storage/storage_with_flush_buffer.rs
```diff
@@ -551,4 +566,27 @@ mod tests {
             FlushWorkers::WORKER_COUNT
         );
     }
+
+    mockall::mock! {
+        pub Storage {}
```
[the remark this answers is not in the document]
psalz: No strong opinion 🤷

psalz (dismissed):
I wish we had static reflection :(

PR #134, opened 2025-09-30
Refactor tests

LorenzSchueler:
This PR refactors tests, mostly in the `storage` module.
The changes are for the most part:
- use `mockall::predicate::eq` instead of writing closures by hand
- use `TestDir` instead of `tempdir`
- impl `AsRef<Path>` for `TestDir`


psalz (approved):

psalz (approved):

ihtefyw (approved):
Good one!

PR #138, opened 2025-10-01
Add NodeSize trait and implemetation for Node, NodeType and NodeId

ihtefyw:
> In preparation of the Weighted Cached Node Manager

Adds a new `NodeSize` trait, which can be used to get information about the byte size and minimum size of a Node type.
This will be used in subsequent PR to calculate the weight of nodes in the CachedNodeManager.

In addition to Node, the trait is also implemented for `NodeType` and `NodeId`. The former allows querying a node size without instantiating a Node, while the latter is useful to retrieve the size of the pointed node without having access to the node itself.

### DIFF
--- a/rust/src/types/node.rs
+++ b/rust/src/types/node.rs
@@ -153,12 +153,12 @@ impl NodeSize for NodeType {
                 std::mem::size_of::<Box<FullLeafNode>>() + std::mem::size_of::<FullLeafNode>()
             }
         };
-        std::mem::size_of::<Self>() + inner_size
+        std::mem::size_of::<Node>() + inner_size
     }
 
     fn min_size() -> usize {
         // Because we don't store empty nodes, the minimum size is the smallest non-empty node.
-        size_of::<Self>() + size_of::<Box<SparseLeafNode<2>>>() + size_of::<SparseLeafNode<2>>()
+        size_of::<Node>() + size_of::<Box<SparseLeafNode<2>>>() + size_of::<SparseLeafNode<2>>()
     }
 }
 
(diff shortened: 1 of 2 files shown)


LorenzSchueler (changes requested):
Please move the actual implementation of `NodeSize` to `NodeType` and then use `size_of` there, and in the implementations for `Node` and `NodeId` use `.to_node_type().byte_size()`. This removes the need for the `LazyLock` setup. And since `min_size` already uses hard coded types, there is no advantage gained in using `size_of_val` for `byte_size` because the types can already get out of sync.

rust/src/types/node.rs
```diff
@@ -117,6 +138,39 @@ pub enum NodeType {
...
+            }
+            NodeType::Leaf2 => {
+                std::mem::size_of::<Box<SparseLeafNode<2>>>()
+                    + std::mem::size_of::<SparseLeafNode<2>>()
+            }
+            NodeType::Leaf256 => {
+                std::mem::size_of::<Box<FullLeafNode>>() + std::mem::size_of::<FullLeafNode>()
+            }
+        };
+        std::mem::size_of::<Node>() + inner_size
```
LorenzSchueler, line 156: nit: you already imported `size_of` and use it directly in `min_size`, so please remove the `std::mem`.
Or even better, import `std::mem` and use `mem::size_of` everywhere. The convention is usually to not import free standing functions directly, but import their parent module.

LorenzSchueler (approved):

rust/src/types/node.rs
ihtefyw: Sorry I missed that

PR #140, opened 2025-10-01
Add double guard to `get_` methods in `CachedNodeManager`

ihtefyw:
Currently, if a thread requests a node that is being deleted, it may wait on acquiring a lock while the delete operation is in progress, obtaining a guard to a non-existing node.

This PR fixes this by again checking if the element is in the cache *after* acquiring the lock, which at that point cannot be deleted anymore.

Note: this should never happen in practice if the tree algorithm keeps a lock on both the current node and its parent when deleting, but IMO it is nice to be shielded from the possibility of that happening.

### DIFF
--- a/rust/src/node_manager/cached_node_manager.rs
+++ b/rust/src/node_manager/cached_node_manager.rs
@@ -18,7 +18,12 @@ use std::{
 
 use quick_cache::{Lifecycle, UnitWeighter};
 
-use crate::{error::Error, node_manager::NodeManager, storage::Storage, types::Node};
+use crate::{
+    error::Error,
+    node_manager::NodeManager,
+    storage::{self, Storage},
+    types::Node,
+};
 
 /// A wrapper which dereferences to [`Node`] and additionally stores its dirty status,
 /// indicating whether it needs to be flushed to storage.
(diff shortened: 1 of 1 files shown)


rust/src/node_manager/cached_node_manager.rs
```diff
@@ -202,39 +202,43 @@ where
...
-            let pos = self.insert(
-                id,
-                NodeWithMetadata {
-                    node,
-                    is_dirty: false,
-                },
-            )?;
-            Ok(self.nodes[pos].read().unwrap())
+            let guard = self.nodes[pos].read().unwrap();
+            if self.cache.get(&id).is_some() {
```
psalz, line 206:
```suggestion
            // The node may have been deleted in the meantime, check again.
            if self.cache.get(&id).is_some() {
```

WDYT? (Same below).

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -202,39 +202,43 @@ where
...
+                return Ok(guard);
+            }
         }
+        // We didn't return, so we need to reload the node from storage.
```
psalz, line 210:
```suggestion
```

If you add the above comment, this one isn't necessary IMO.

psalz (approved):

PR #141, opened 2025-10-01
Add memory capacity bound to `CachedNodeManager`

ihtefyw:
This PR adds a memory upper bound to the memory used by the `CachedNodeManager` to better control the Carmen memory consumption.

The `CachedNodeManager::new` now gets a **bytes_capacity** parameter, which is used as the available memory budget to allocate all the internal components, i.e. the `nodes` slice and the internal cache. Every item is then weighted against their byte size, which is used to determine how much space they require in the cache.
The current algorithm works as follows:

1. Calculate the footprint of the **minimum size** element in the cache: this is `element_byte_size + sizeof(RwLock<Wrapper>) + cache_overhead_per_item`. The cache overhead formula is derived from [quick_cache docs](https://docs.rs/quick_cache/latest/quick_cache/#approximate-memory-usage). 
2. Allocate `bytes_capacity / minimum_element_footprint` elements: this covers the worst case of only inserting the smallest node in the manager. This would never happen in practice, and it's a quite big approximation.
3. Set the cache weight as `num_nodes * minimum_element_size`: this ensures that, in the worst case, the manager doesn't allocate more than the available slots in the `nodes` vector. 

This has an important drawback: the amount of memory wasted is proportional to the difference between the average size of the nodes stored in the manager and the theoretically minimum size, as the cache will quickly fill up and multiple positions in the `nodes` vector will be unused.

Another drawback is that the number of `quick_cache` shards is overestimated, as it's calculated against the worst case of just minimum sized nodes. This has a direct impact on parallelism and should be benchmarked to find the best balance.

Additional changes:
- The `evict` function now takes a slice, as inserting an element may require to evict multiple elements to make space for its weight.
- The `ItemLifecycle` component now returns a vector, as multiple elements may be evicted during a single operation

Depends on #138 #143 

Closes https://github.com/0xsoniclabs/sonic-admin/issues/382

### DIFF
--- a/rust/src/node_manager/cached_node_manager.rs
+++ b/rust/src/node_manager/cached_node_manager.rs
@@ -11,24 +11,25 @@
 use std::{
     cmp::Eq,
     hash::{Hash, RandomState},
+    mem,
     ops::{Deref, DerefMut},
     sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
 
 use dashmap::DashSet;
-use quick_cache::{Lifecycle, UnitWeighter};
+use quick_cache::{Lifecycle, Weighter};
 
 use crate::{
     error::Error,
     node_manager::NodeManager,
     storage::{Checkpointable, Storage},
-    types::Node,
+    types::{Node, NodeSize},
 };
 
 /// A wrapper which dereferences to [`Node`] and additionally stores its dirty status,
 /// indicating whether it needs to be flushed to storage.
 /// The node's status is set to dirty when a mutable reference is requested.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct NodeWithMetadata<N> {
     node: N,
     is_dirty: bool,
@@ -49,6 +50,10 @@ impl<N> DerefMut for NodeWithMetadata<N> {
     }
 }
 
+/// A node manager that caches nodes in memory, with a underlying storage backend.
+/// Nodes are retrieved from the underlying storage if they are not present in the cache, and store
+/// them back when evicted if they have been modified.
+/// It only guarantees that no access to a deleted node is possible.
 pub struct CachedNodeManager<K, N, S> {
     // A fixed-size container that acts as the owner of all nodes.
     // This allows us to hand out read/write guards from a shared reference to the node manager.
@@ -58,7 +63,7 @@ pub struct CachedNodeManager<K, N, S> {
     cache: quick_cache::sync::Cache<
         K,                // key type to identify cached values
         usize,            // value type representing slots in `Self::nodes`
-        UnitWeighter,     // all values are considered to cost the same
+        ByteSizeWeighter, // all values are considered to cost the same
         RandomState,      // default hasher
         ItemLifecycle<N>, // tracks and reports evicted items
     >,
@@ -67,17 +72,31 @@ pub struct CachedNodeManager<K, N, S> {
     storage: S,
 }
 
-impl<K: Eq + Hash + Copy, S, N> CachedNodeManager<K, N, S>
+impl<K, S, N> CachedNodeManager<K, N, S>
 where
     S: Storage<Id = K, Item = N>,
-    N: Default,
+    N: Default + NodeSize,
+    K: Eq + Hash + Copy + NodeSize,
 {
-    /// Creates a new [`CachedNodeManager`] with the given capacity and storage backend.
-    pub fn new(capacity: usize, storage: S) -> Self {
+    /// Creates a new [`CachedNodeManager`] with the given capacity (in bytes) and storage backend.
+    /// NOTE: the capacity must be at least 1MB, or in general enough to store at least two nodes.
+    pub fn new(bytes_capacity: usize, storage: S) -> Self {
+        // Requires at least 1MB
+        if bytes_capacity < 1024 * 1024 {
+            panic!("Node manager capacity too small. Please provide a larger capacity.");
+        }
+        // NOTE: Here we could just take into consideration the element footprint and ignore the
+        // cache overhead. This would increase the number of nodes we can store and increase the
+        // maximum cache weight capacity.
+        let min_node_byte_size =
+            N::min_non_empty_node_size() + Self::get_element_memory_footprint();
         // We allocate a slot for one additional node. This way, when the cache is full, we always
         // have a free slot we can use to insert a new item into the cache and force the
         // eviction of an old one.
-        let num_nodes = capacity + 1;
+        let num_nodes = bytes_capacity / min_node_byte_size + 1; // + 1 to be sure there is always gonna be a free node available
+        // NOTE: cache_weight needs to be <= num_nodes * N::min_non_empty_node_size() to ensure that
+        // the cache weight is exhausted when inserting only nodes with the minimum size.
+        let cache_weight = (num_nodes - 1) * N::min_non_empty_node_size();
         let nodes: Arc<[_]> = (0..num_nodes)
             .map(|_| {
                 // Pre-allocate with default values. This requires `N: Default`.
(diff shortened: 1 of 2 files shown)


rust/src/node_manager/cached_node_manager.rs
```diff
@@ -108,28 +128,29 @@ where
...
+            **guard = N::default(); // reset node to default value to release storage
+            self.free_slots.insert(*pos);
         }
-        **guard = N::default(); // reset node to default value to release storage
-        self.free_slots.insert(pos);
         Ok(())
     }
 
     /// Insert an item in the node manager, reusing a free slot if available or evicting an
     /// existing item if the cache is full.
```
LorenzSchueler, line 148:
```suggestion
/// Insert an item in the node manager, reusing a free slot if available or evicting
   /// existing items if the cache is full.
```

rust/src/node_manager/cached_node_manager.rs
LorenzSchueler, line 161:
```suggestion
        let pos = loop {
            let pos = self.free_slots.iter().next().map(|p| *p);
            if let Some(pos) = pos
                && let Some(pos) = self.free_slots.remove(&pos)
            {
                break pos;
            }
            std::hint::spin_loop();
        };
```

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -141,21 +162,26 @@ where
...
-        }
+        self.on_evict(evicted.as_slice())?;
         Ok(pos)
     }
+
+    /// Returns the memory required to store a node, including wrappers and cache overhead.
+    /// For the cache overhead, see [quick_cache docs](https://docs.rs/quick_cache/latest/quick_cache/#approximate-memory-usage)
+    pub fn get_element_memory_footprint() -> usize {
+        mem::size_of::<RwLock<NodeWithMetadata<N>>>() // Stored wrapper in the nodes vector
+    + (((mem::size_of::<K>() + mem::size_of::<usize>() + 21) as f64 * 1.5).floor() as usize).next_power_of_two() // Cache overhead per item
```
LorenzSchueler, line 176:
```suggestion
        let wrapper_size = mem::size_of::<RwLock<NodeWithMetadata<N>>>();
        let cache_overhead = (((mem::size_of::<K>() + mem::size_of::<usize>() + 21) as f64 * 1.5)
            .floor() as usize)
            .next_power_of_two();
        wrapper_size + cache_overhead
```

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -304,16 +347,110 @@ mod tests {
...
+                + ((size_of::<NodeId>() + 21 + size_of::<usize>())
+                    * ((manager.nodes.len() - 1) as f64 * 1.5) as usize)
+                    .next_power_of_two()
+                <= capacity
+        );
+    }
+
+    #[test]
+    #[should_panic]
+    fn cached_node_manager_new_fails_if_capacity_too_small() {
```
LorenzSchueler, line 372:
```suggestion
    fn cached_node_manager_new_panics_if_capacity_too_small() {
```
We use `fails` often times when we mean that it returns an error.

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -304,16 +347,110 @@ mod tests {
...
+        ];
+
+        for case in cases {
+            let mut storage = MockCachedNodeManagerStorage::new();
+            storage.expect_set().returning(move |_, _| Ok(()));
+            let num_nodes = cache_capacity
+                / (case
+                    .iter()
+                    .map(|c| c.node_byte_size() / case.len())
+                    .sum::<usize>());
```
LorenzSchueler, line 395:
```suggestion
            let num_nodes = cache_capacity
                / (case.iter().map(NodeSize::node_byte_size).sum::<usize>() / case.len());
```

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -304,16 +347,110 @@ mod tests {
...
+    #[test]
+    fn cached_node_manager_does_not_exceed_capacity() {
+        let max_capacity = ONE_MB;
+        let cache_capacity = get_cache_capacity(max_capacity) as usize;
+        let cases = [
+            vec![NodeType::Leaf2],
+            vec![NodeType::Leaf256],
+            vec![NodeType::Inner],
+            vec![NodeType::Leaf2, NodeType::Leaf256, NodeType::Inner],
+        ];
```
LorenzSchueler, line 386: use rstest?

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -304,16 +347,110 @@ mod tests {
...
+                / (case
+                    .iter()
+                    .map(|c| c.node_byte_size() / case.len())
+                    .sum::<usize>());
+            let manager = TestCachedNodeManager::new(max_capacity, storage);
+
+            for i in 0..(num_nodes * 2) {
+                let node_type = case[i % case.len()];
+                let id = NodeId::from_idx_and_node_type(i as u64, node_type);
+                let _ = manager
```
LorenzSchueler, line 401: nit:
```suggestion
                manager
```

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -304,16 +347,110 @@ mod tests {
...
+            let cache_elements_weight = manager
+                .cache
+                .iter()
+                .map(|(id, _)| {
+                    id.node_byte_size() + TestCachedNodeManager::get_element_memory_footprint()
+                })
+                .sum::<usize>();
+            // Size of unused elements in the node vector
+            let vector_overhead = (manager.nodes.len() - manager.cache.len())
+                * (size_of::<RwLock<NodeWithMetadata<Node>>>() + size_of::<Box<Node>>());
```
LorenzSchueler, line 426: Where is `Box<Node>` coming from?

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -304,16 +347,110 @@ mod tests {
...
+                    id,
+                    NodeWithMetadata {
+                        node: Node::Leaf2(Box::default()),
+                        is_dirty: true,
+                    },
+                )
+                .unwrap();
+        }
+        assert!(
+            !manager.free_slots.is_empty(),
```
LorenzSchueler, line 452: This is testing that free slots is not empty, but not that it is not full.

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -674,4 +848,11 @@ mod tests {
         Ok((**guard).clone())
     }) as GetMethod)]
     fn get_method(#[case] f: GetMethod) {}
+
+    /// Helper function to get the actual cache capacity for a given byte size.
+    fn get_cache_capacity(size: usize) -> u64 {
```
LorenzSchueler, line 853:
```suggestion
    fn get_cache_weight_capacity(size: usize) -> u64 {
```

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -469,80 +598,103 @@ mod tests {
...
-        for i in 0..NUM_NODES + 1 {
-            // 1 item more than capacity
+        let max_capacity = ONE_MB;
+        let cache_capacity = get_cache_capacity(max_capacity) as usize;
+        let cases = [
+            vec![NodeType::Leaf2],
+            vec![NodeType::Leaf256],
+            vec![NodeType::Inner],
+            vec![NodeType::Leaf2, NodeType::Leaf256, NodeType::Inner],
+        ];
```
LorenzSchueler, line 608: use rstest?

LorenzSchueler (changes requested):

ihtefyw:
This PR is very outdated.
It has been verified that with Verkle Tries the `CachedNodeManager` size does not currently influence the performance significantly. Moreover, the current size of 1.000.000 nodes results in an average size of 1/2GB (worst, but impossible, case 8GB).
Therefore, the complexity introduced by this PR are not justified anymore.

I am closing this for now, in case this results into an issue with other data structure implementations we can reopen it.

PR #143, opened 2025-10-02
Switch to `DashSet` in cached node manager + remove `PINNED_POS`

ihtefyw:
This PR substitutes the free_list vector in cached_node_manager with `DashSet`, which allows for more fine-grained synchronization.
Moreover, the node manager now allocates an additional node to make sure that there will always be a free position when inserting. This removes the need of the `PINNED_POS` special key as an element will never have to be inserted twice in the cache.

### DIFF
--- a/rust/src/node_manager/cached_node_manager.rs
+++ b/rust/src/node_manager/cached_node_manager.rs
@@ -12,10 +12,10 @@ use std::{
     cmp::Eq,
     hash::{Hash, RandomState},
     ops::{Deref, DerefMut},
-    sync::{Arc, Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard},
-    vec::Vec,
+    sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
 
+use dashmap::DashSet;
 use quick_cache::{Lifecycle, UnitWeighter};
 
 use crate::{
@@ -54,15 +54,15 @@ pub struct CachedNodeManager<K, N, S> {
     // This allows us to hand out read/write guards from a shared reference to the node manager.
     // Wrapped in an Arc so that it can be shared with [`ItemLifecycle`].
     nodes: Arc<[RwLock<NodeWithMetadata<N>>]>,
-    // cache, managing the key to node position mapping as well as the item eviction
+    /// The cache managing the key-to-slot mapping as well as item eviction
     cache: quick_cache::sync::Cache<
         K,                // key type to identify cached values
-        usize,            // value type to identify node positions in the nodes vector
+        usize,            // value type representing slots in `Self::nodes`
         UnitWeighter,     // all values are considered to cost the same
         RandomState,      // default hasher
         ItemLifecycle<N>, // tracks and reports evicted items
     >,
-    free_list: Mutex<Vec<usize>>, // free list of available node positions
+    free_slots: DashSet<usize>, // set of free slots in [`Self::nodes`]
     //storage for managing IDs, fetching missing nodes, and saving evicted nodes to
     storage: S,
 }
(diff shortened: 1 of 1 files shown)


rust/src/node_manager/cached_node_manager.rs
```diff
@@ -126,53 +117,34 @@ where
...
-            // cache eviction failed (e.g. since all items are pinned) and
-            // the insertion cannot proceed.
-            // Note: this should never happen as we checked before that the eviction returned an
-            // item.
-            free_list_guard.pop().ok_or(Error::NodeManager(
-                "no available space in cache".to_string(),
-            ))?
+        // While there should always be at least one free position available, there is a interval in
+        // which the list may be empty while inserting a new item and evicting an old one.
+        // In that case, we loop until a free position is available.
```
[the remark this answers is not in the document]
ihtefyw: It would require quick_cache to use shuttle's primitives

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -607,10 +559,14 @@ mod tests {
 
         let manager = CachedNodeManager::new(2, storage);
         let _ = manager.add(entry).unwrap();
+        let _ = manager.cache.get(&id).expect("entry should be in cache");
         manager.delete(id).unwrap();
         assert!(manager.cache.get(&id).is_none());
         // First node should be inserted at pos 0
```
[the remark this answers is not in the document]
ihtefyw: Redundant

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -528,11 +479,10 @@ mod tests {
         }
         storage
             .expect_set()
-            .times(1)
             .with(
-                eq(NodeId::from_idx_and_node_type(
+                ne(NodeId::from_idx_and_node_type(
                     // NUM_ELEMENT - 1 will be evicted because of infinite reuse distance
-                    NUM_NODES - 1,
+                    NUM_NODES,
```
psalz, line 485:
```suggestion
                    // The NUM_NODES-th node will be evicted because of infinite reuse distance
                    NUM_NODES,
```

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -248,26 +220,22 @@ where
...
-            if pos == ItemLifecycle::<N>::PINNED_POS {
-                continue; // skip pinned items
-            }
             let mut entry_guard = self.nodes[pos].write().unwrap();
-            // Skip deleted nodes. We expect the free list to be short, so this should be cheap.
+            // Skip unused slots. We expect the free list to be short, so this should be cheap.
             // NOTE: this operation may be quadratic in the number of nodes. However,
             // we expect the cache to be always full, therefore the cost of this op should amortize
             // quickly.
             // TODO: Benchmark this and see if it is a problem in practice.
```
psalz, line 237:
```suggestion
```

psalz (changes requested):

LorenzSchueler (approved):

rust/src/node_manager/cached_node_manager.rs
```diff
@@ -128,53 +119,34 @@ where
...
-            // cache eviction failed (e.g. since all items are pinned) and
-            // the insertion cannot proceed.
-            // Note: this should never happen as we checked before that the eviction returned an
-            // item.
-            free_list_guard.pop().ok_or(Error::NodeManager(
-                "no available space in cache".to_string(),
-            ))?
+        // While there should always be at least one free position available, there is a interval in
+        // which the list may be empty while inserting a new item and evicting an old one.
+        // In that case, we loop until a free position is available.
```
psalz, line 132:
```suggestion
        // While there should always be at least one free slot available, there is a interval in
        // which the set may be empty while another thread is inserting a new item and before it has evicted an old one.
        // In that case, we loop until a free position is available.
```

PR #144, opened 2025-10-02
Register Rust in-memory S6 implementation in Go

psalz:
This is the last piece of the puzzle required to call into the Rust in-memory S6 implementation from the Go side. From now on, in addition to the Rust-side tests, all changes made to S6 will also be checked by the Go test suite.

Thankfully #80 got us basically 99% of the way to making the Go tests compatible with S6, the only thing remaining is to add a workaround for a test of the `GetMemoryFootprint` function, which is not yet implemented in Rust.

Depends on #130.
Closes https://github.com/0xsoniclabs/sonic-admin/issues/272.

### DIFF
--- a/go/state/externalstate/configurations.go
+++ b/go/state/externalstate/configurations.go
@@ -16,6 +16,8 @@ const (
 	VariantCppMemory  state.Variant = "cpp-memory"
 	VariantCppFile    state.Variant = "cpp-file"
 	VariantCppLevelDb state.Variant = "cpp-ldb"
+
+	VariantRustMemory state.Variant = "rust-memory"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


kjezek (approved):
looks good, thanks

go/state/externalstate/configurations.go
```diff
@@ -45,4 +45,10 @@ func init() {
 			}, newCppLevelDbBasedState)
 		}
 	}
+
+	state.RegisterStateFactory(state.Configuration{
+		Variant: "rust-memory",
```
LorenzSchueler, line 50: nit: Consider using a consts like the ones for the C++ implementations.

LorenzSchueler (approved):

PR #147, opened 2025-10-07
Use `shuttle` to find thread interleaving bugs in `CachedNodeManager`

ihtefyw:
The `CachedNodeManager` gives a certain number of guarantees when used in a multithreaded environment, which are not currently checked.

This PR introduces a series of tests using a tool called [shuttle](https://github.com/awslabs/shuttle), which allows trying multiple thread interleaving in a test, allowing to check if certain conditions hold. 
I added two classes of tests to the `CachedNodeManager`:
- `shuttle_*_never_returns_a_reference_to_a_non_existing_node_on_*`
- `shuttle_*_always_terminate`

Shuttle requires switching all synchronization primitives to their implementation to work. Therefore, I added a new `sync` file with all the synchronization primitives and a compile time check to select the desired one (the std are enabled by default).
Because running tests outside the shuttle context with the shuttle primitives is not possible, shuttle needs to be explicitly enable through the `--features shuttle` option and filter tests bases on the `shuttle_` prefix.

Depends on #143 #141


rust/Cargo.toml
```diff
@@ -29,7 +29,7 @@ harness = false
 [dependencies]
 # datastructures
 dashmap = { version = "6.1.0", features = ["inline"] }
-quick_cache = "0.6.15"
+quick_cache = { version = "0.6.1" }
```
LorenzSchueler, line 32: revert this

rust/Cargo.toml
```diff
@@ -45,6 +45,12 @@ ipa-multipoint = { git = "https://github.com/crate-crypto/rust-verkle.git", rev
 verkle-trie = { git = "https://github.com/crate-crypto/rust-verkle.git", rev = "e27b8b4" }
 sha3 = "0.10.8"
 
+# testing
+shuttle = { version = "0.8.1", optional = true }
```
LorenzSchueler, line 49: please make it a dev dep

rust/src/sync.rs
```diff
@@ -0,0 +1,32 @@
...
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+#[cfg(not(feature = "shuttle"))]
+pub(crate) use std::sync::{
+    Arc, Mutex, RwLock,
+    atomic::{AtomicBool, AtomicU64, Ordering},
+};
+#[cfg(not(feature = "shuttle"))]
+#[cfg(test)]
```
LorenzSchueler, line 17:
```suggestion
#[cfg(all(test, not(feature = "shuttle")))]
```

LorenzSchueler (commented):
This is very useful 👍🏽. However, I have one concern: `shuttle` mirrors the `std::sync` API. This means we are locking ourselves in and can not switch to e.g. `parking_lot` in the future, which has a different API.
Below, a few comments. I have not yet looked at the tests themselves, because I think you are still working on them.

PR #342, opened 2026-02-06
Add Synchronous Archive updates.

facuMH:
This PR introduces the ability to apply block changes to the live state while ensuring the operation waits until the archive state has also applied those changes. This is critical for processes that require strict consistency between the live state and the archive before proceeding.

#### Key Contributions
1. `StateDB.EndBlockSync`
Introduces `EndBlockSync` to the `StateDB` interface. This method shares the logic of the existing `EndBlock` but utilizes the internal `State.ApplySync` newly introduced function to ensure synchronous completion across different state layers.

2. `State.ApplySync` Interface Expansion
Added `ApplySync` to the State interface.

#### Specialized Behaviour
While many implementations may default to standard `Apply` logic, this method allows for implementations to use blocking wait-logic for archival confirmation when needed.

#### Testing
Added unit tests for the two implementations where `ApplySync` introduces distinct behaviour compared to the standard `Apply` method.

- [x] Verified that the synchronous wait-cycle correctly holds the caller until the archive state is updated.
- [x] Verified that synchronous and asynchronous can be interleaved and order is maintained.
- [x] Verified that synchronous calls wait for archive to be written.
- [x] Verified that asynchronous calls do not wait for archive to be written but trigger it.

### DIFF
--- a/go/database/vt/reference/state.go
+++ b/go/database/vt/reference/state.go
@@ -171,6 +171,11 @@ func (s *State) Apply(block uint64, update common.Update) error {
 	return nil
 }
 
+func (s *State) ApplySync(block uint64, update common.Update) error {
+	// Apply is already synchronous
+	return s.Apply(block, update)
+}
+
 func (s *State) GetHash() (common.Hash, error) {
 	return s.GetCommitment().Await().Get()
 }
--- a/go/state/gostate/go_state.go
+++ b/go/state/gostate/go_state.go
@@ -41,9 +41,24 @@ type GoState struct {
 	archiveWriterFlushDone <-chan bool
 	archiveWriterDone      <-chan bool
 	archiveWriterError     <-chan error
+
+	// testingPing testing only channel
+	testingPing chan testingPingOrigin
 }
 
+// testingPingOrigin is an enum used for testing to identify the source of a ping on the testingDone channel.
+type testingPingOrigin int8
+
+var (
+	fromUpdate testingPingOrigin = 0
+	fromApply  testingPingOrigin = 1
+)
+
 func newGoState(live state.LiveDB, archive archive.Archive, cleanup []func()) state.State {
+	return state.WrapIntoSyncedState(_newGoState(live, archive, cleanup))
+}
+
+func _newGoState(live state.LiveDB, archive archive.Archive, cleanup []func()) *GoState {
 
 	res := &GoState{
 		live:    live,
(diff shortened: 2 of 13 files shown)


go/state/gostate/go_state.go
```diff
@@ -41,9 +41,24 @@ type GoState struct {
 	archiveWriterFlushDone <-chan bool
 	archiveWriterDone      <-chan bool
 	archiveWriterError     <-chan error
+
+	// testingPing testing only channel
+	testingPing chan testingPingOrigin
 }
 
+// testingPingOrigin is an enum used for testing to identify the source of a ping on the testingDone channel.
```
LorenzSchueler:
```suggestion
// testingPingOrigin is an enum used for testing to identify the source of a ping on the testingPing channel.
```

LorenzSchueler (commented):
Having the `testingPing` channel looks a bit cumbersome at first, but from the approaches we discusses do far, I think it allows for the best tests.
For all tests that are only for ApplySync xor Apply, it would be nice if we would run them for both anyway, and simply expect that one fails (that is the test should be reported as successful iff the actual test execution failed). Not sure if Go supports this.

go/state/gostate/go_state_test.go
```diff
@@ -769,6 +770,126 @@ func TestState_Apply_CannotCallRepeatedly_OnError(t *testing.T) {
 	}
 }
 
+func TestState_Apply_DoesNotWaitForArchiveWriteToFinish(t *testing.T) {
```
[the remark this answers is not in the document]
facuMH: it can be done, but I think keeping them separate helps readability.

go/database/vt/reference/state.go
```diff
@@ -171,6 +171,11 @@ func (s *State) Apply(block uint64, update common.Update) error {
 	return nil
 }
 
+func (s *State) ApplySync(block uint64, update common.Update) error {
+	// Apply is already synchronous
```
LuisPH3: this comment does not really explain anything.

go/state/state.go
```diff
@@ -56,6 +56,11 @@ type State interface {
 	// Apply applies the provided updates to the state content.
 	Apply(block uint64, update common.Update) error
 
+	// ApplySync applies the provided updates to the state content synchronously.
+	// Each implementation may have a different mechanism, but all must ensure the
+	// updates are fully applied when the method returns.
```
HerbertJordan:
```suggestion
	// ApplySync applies the provided updates to the state content. Unlike the
	// Apply function, ApplySync blocks until the archive state is also in sync with
	// the live state.
```

go/state/gostate/go_state.go
```diff
@@ -99,6 +121,7 @@ type archiveUpdate = struct {
 	block       uint64
 	update      *common.Update  // nil to signal a flush
 	updateHints common.Releaser // an optional field for passing update hints from the LiveDB to the Archive
+	sync        chan struct{}   // an optional channel which gets closed when the update was processed
```
LuisPH3: nit: write only channel on this side.

LuisPH3 (commented):
This PR is too large, please break it down. 
There is a lot of "fluff" but review is very uncomfortable at this size. Please break it down into more manageable units.

facuMH:
A different approach will be used to implement this feature.

PR #370, opened 2026-03-03
Add `withinTransaction` field and make `undo` multi dimensional

ihtefyw:
This PR does a set of changes in the `stateDB` required for the implementation of bundle transaction support.
In order:
- Adds an explicit `withinTransaction` state, and adds guards to Begin/End Transaction and EndBlock. BeginBlock is ignored as it's a no-op.
- Makes the `state_db.undo` multi-dimensional, with one entry per transaction, and an undo vector for each transaction. This will help the implementation of the transaction-level revert functionality.


LorenzSchueler (changes requested):
I think the checks can be improved a little bit more. See comments below.

PR #393, opened 2026-04-20
Add live-archive synchronization integration test

ihtefyw:
This PR adds a simple integration test to verify the live and archive db synchronization after an update.
The only way to test it that I found was to simply wait on the returned channel until is closed and check if the data in the archive are present.
Other options considered:
- `synctest` is not an option as the way the archive update is implemented depends on the state implementation, and we can't assume there is an asynchronicity with blocked goroutines
- A select statement with an aribitrary timer would put constraints on the time an archive update can take. We could put a long timer (e.g. 1 or 10 seconds), but anyway the go tests have timers, and it's gonna fail anyway

### DIFF
--- a/go/state/state_db_and_state_test.go
+++ b/go/state/state_db_and_state_test.go
@@ -850,6 +850,49 @@ func TestStateDB_CallingExistsAfterAccountIsDeletedReturnsFalse(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


HerbertJordan (approved):

facuMH (approved):
LGTM

PR #402, opened 2026-05-05
Split Rust and C++ externastate and make them optional

ihtefyw:
Currently, the C++ and Rust implementation are required to run the Go tests. Moreover, when including the `externalstate`, it is not possible to use the Rust implementation without building the C++ one too, and vice versa.

This PR does two things:
- Makes the `externastate` package optional in tests
- Splits the C++ and Rust implementation, allowing to pick which implementation to include in test/build without having to build the other one.

It is structured as follows: each implementation has a `connector_IMPL.go`, which defines the bindings and the state factory initialization. Each connector has a build tag guard called `carmen_IMPL`.   
The `external_state.go` file is enabled conditionally if any of the build tags for rust or cpp is provided.


go/state/externalstate/configurations.go
HerbertJordan, line 1: If I recall correctly, you can have multiple `init` functions in the same package. So this file should not be necessary. You can rename `initCpp` and `initRust` both to `init` and be fine. Also, no defaults needed in this case.

HerbertJordan (changes requested):
I think the setup can be simplified by exploiting the fact that you can have multiple "init" functions in the same package.

go/state/externalstate/configurations.go
ihtefyw: thanks, that removed a lot of complexity!

LorenzSchueler (approved):
Nice

go/state/externalstate/rust_connector.go
```diff
@@ -0,0 +1,160 @@
...
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+//go:build carmen_rust
+
+package externalstate
+
+/*
```
LorenzSchueler, line 17: We can also use go generate to build the rust code, similar to how we call the shell script to build c++. Please also document this in the readme
```suggestion
package externalstate

//go:generate cargo build --release --manifest-path=../../../rust/Cargo.toml

/*
```

LorenzSchueler (changes requested):

HerbertJordan (dismissed):

LorenzSchueler (dismissed):

ihtefyw:
@HerbertJordan I had to introduce again a `default.go` file to avoid having an empty package when none of the bindings are enabled.
I also added a CI check as the code should compile with each combination of tags

LorenzSchueler (approved):

PR #405, opened 2026-05-11
Cherry Pick: Test that externalstate configurations are added to global list (#404)

ihtefyw:
This PR cherry picks e29019e070097c6f0822223fe6a6e3340270bab9 in `sonic_2.2`.
In particular, it removes the test `TestConfigurations_ContainGoAndCppImplementations`, which was failing and replace it with two more fine-grained tests

### DIFF
--- a/go/state/externalstate/cpp_connector.go
+++ b/go/state/externalstate/cpp_connector.go
@@ -23,6 +23,7 @@ package externalstate
 */
 import "C"
 import (
+	"maps"
 	"unsafe"
 
 	"github.com/0xsoniclabs/carmen/go/database/flat"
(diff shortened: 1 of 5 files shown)


go/state/externalstate/cpp_connector.go
```diff
@@ -152,13 +153,19 @@ func init() {
 		}
 	}
 
-	// Register all experimental configurations.
+	flatFactories := map[state.Configuration]state.StateFactory{}
```
LorenzSchueler, line 156: Why not add it directly to `factories` if you copy them over later on anyway?

go/state/externalstate/rust_connector.go
```diff
@@ -150,13 +151,19 @@ func init() {
 		Archive: "file",
 	}] = newRustFileBasedState
 
-	// Register all experimental configurations.
+	flatFactories := map[state.Configuration]state.StateFactory{}
```
LorenzSchueler, line 154: as above

LorenzSchueler (approved):

HerbertJordan (approved):

go/state/externalstate/cpp_connector.go
ihtefyw: Because you're iterating over them, and it would add the same factories multiple time to the range you're iterating on

PR #415, opened 2026-05-18
Show that `CreateAccount` and `Suicide` do not alter `State` in some conditions

ihtefyw:
This PR adds to tests for `CreateAccount` and `Suicide` to show how, under the current rules, they don't have any impact on the `State`.

- `TestStateDB_AccountsAreCreatedImplicitlyWhenSettingState` shows that accounts are always created implicitly, and the call to `CreateAccount` is superfluous
- `TestStateDB_DeleteAccountCreatedInSameTransactionDoesNotReachState` shows that deleting an account created in the current transaction is a no-op on the `State`. This is the only way to call `Suicide`.

This, in addition to the comments in https://github.com/0xsoniclabs/sonic-admin/issues/762, is to show that `CreateAccount` and `DeleteAccount` fields in Updates can be safely removed. 
Following PRs to remove those functions with the related tests will follow

### DIFF
--- a/go/state/state_db_and_state_test.go
+++ b/go/state/state_db_and_state_test.go
@@ -914,12 +914,12 @@ func executeOpOnClearStateDB(t *testing.T, config namedStateConfig, op func(stat
 }
 
 func TestStateDB_AccountsAreCreatedImplicitlyWhenSettingState(t *testing.T) {
-	funcList := map[string]func(s state.StateDB){
-		"SetState": func(s state.StateDB) {
-			s.SetState(address1, key1, val1)
-		},
+	ops := map[string]func(s state.StateDB){
 		"AddBalance": func(s state.StateDB) {
 			s.AddBalance(address1, amount.New(10))
+		},
+		"SetState": func(s state.StateDB) {
+			s.SetState(address1, key1, val1)
 		},
 		"SetNonce": func(s state.StateDB) {
 			s.SetNonce(address1, 10)
@@ -933,7 +933,7 @@ func TestStateDB_AccountsAreCreatedImplicitlyWhenSettingState(t *testing.T) {
 		if config.config.Schema < 4 {
 			continue
 		}
-		for name, fn := range funcList {
+		for name, fn := range ops {
 			t.Run(config.name()+"_"+name, func(t *testing.T) {
 				t.Parallel()
 				require := require.New(t)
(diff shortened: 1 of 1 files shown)


go/state/state_db_and_state_test.go
```diff
@@ -893,6 +893,84 @@ func TestStateDB_ArchiveIsSynchronizedWithLiveDB(t *testing.T) {
 	}
 }
 
+// executeOpOnClearStateDB executes the given operation on a clear stateDB in single transaction/block and returns the state root hash.
+func executeOpOnClearStateDB(t *testing.T, config namedStateConfig, op func(state.StateDB)) common.Hash {
```
LorenzSchueler:
```suggestion
// executeOpOnClearStateDB executes the given operation on a clean stateDB in single transaction/block and returns the state root hash.
func executeOpOnCleanStateDB(t *testing.T, config namedStateConfig, op func(state.StateDB)) common.Hash {
```

go/state/state_db_and_state_test.go
```diff
@@ -893,6 +893,84 @@ func TestStateDB_ArchiveIsSynchronizedWithLiveDB(t *testing.T) {
...
+		},
+		"SetState": func(s state.StateDB) {
+			s.SetState(address1, key1, val1)
+		},
+		"SetNonce": func(s state.StateDB) {
+			s.SetNonce(address1, 10)
+		},
+		"SetCode": func(s state.StateDB) {
+			s.SetCode(address1, []byte{0xAC})
+		},
```
LorenzSchueler, line 933: What about `SubBalance` and `SetTransientState`?

ihtefyw: SubBalance is missing, gonna add it.
SetTransientState is local to each transaction, it's not used to flush anything to the underlying `State`

LorenzSchueler (dismissed):

go/state/state_db_and_state_test.go
```diff
@@ -893,6 +893,88 @@ func TestStateDB_ArchiveIsSynchronizedWithLiveDB(t *testing.T) {
...
+			continue
+		}
+		t.Run(config.name(), func(t *testing.T) {
+			t.Parallel()
+			require := require.New(t)
+			with := executeOpOnCleanStateDB(t, config, func(s state.StateDB) {
+				s.CreateAccount(address1)
+				s.Suicide(address1)
+			})
+			without := executeOpOnCleanStateDB(t, config, func(s state.StateDB) {
```
HerbertJordan: one more case I would be curious about - an implicit create, as this will be the main way of creating an account after CreateAccount is removed:
```
			with := executeOpOnCleanStateDB(t, config, func(s state.StateDB) {
				s.SetNonce(address1, 1)
				s.Suicide(address1)
			})
```

HerbertJordan (dismissed):
Thanks for testing this out, looks good to me.

LorenzSchueler (approved):

PR #417, opened 2026-05-20
Remove `CreateAccount` from `State` implementations

ihtefyw:
This PR removes the `CreateAccount` function from the `UpdateTarget` interface, and therefore from each `State` implementation.
On his own, `CreateAccount` could never be called, and was always followed by a state-changing operation (balance, nonce, code).

This also removes the empty code hash set by the state implementations when creating an account: this was complicating the implementation, adding an unexpected side effect to the update functions. 
A note: the `geth-leveldb` VK implementation uses the leveldb geth implementation, which sets all the fields in the account update, including the code hash, in each update. Therefore, it is not possible not to set the empty code hash in this implementation. For such reason, tests changing only one value and then checking the resulting hash are skipped for this implementation.

Note that the CreateAccount function is still available in the `VMStateDB` interface, where it's called from the EVM.

### DIFF
--- a/go/common/update.go
+++ b/go/common/update.go
@@ -23,7 +23,7 @@ import (
 //go:generate mockgen -source update.go -destination update_mocks.go -package common
 
 // Update summarizes the effective changes to the state DB at the end of a block.
-// It combines changes to the account state (created or deleted), balances, nonces
+// It combines changes to the account state (deleted), balances, nonces
 // codes, and slot updates.
 //
 // An example use of an update would look like this:
@@ -32,8 +32,8 @@ import (
 //	update := Update{}
 //	// Fill in changes.
 //	// Note: for each type of change, updates must be in order and unique.
-//	update.AppendCreateAccount(..)
-//	update.AppendCreateAccount(..)
+//	update.AppendBalanceUpdate(..)
+//	update.AppendBalanceUpdate(..)
 //	...
 //	// Optionally, check that the provided data is valid (sorted and unique).
 //	err := update.Check()
@@ -41,7 +41,6 @@ import (
 // Valid instances can then be forwarded to the State as a block update.
 type Update struct {
 	DeletedAccounts []Address
-	CreatedAccounts []Address
 	Balances        []BalanceUpdate
 	Nonces          []NonceUpdate
 	Codes           []CodeUpdate
@@ -51,7 +50,6 @@ type Update struct {
 // IsEmpty is true if there is no change covered by this update.
 func (u *Update) IsEmpty() bool {
 	return len(u.DeletedAccounts) == 0 &&
-		len(u.CreatedAccounts) == 0 &&
 		len(u.Balances) == 0 &&
 		len(u.Nonces) == 0 &&
 		len(u.Codes) == 0 &&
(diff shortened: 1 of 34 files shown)


go/state/state_db.go
```diff
@@ -1288,7 +1288,6 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
 	for addr, value := range s.accounts {
 		if value.original != value.current {
 			if value.current == accountExists {
```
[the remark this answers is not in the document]
ihtefyw: Only contracts can be created and deleted, but only in the same tx. This means nothing should reach the underlying state.

go/common/update.go
```diff
@@ -442,11 +393,23 @@ func (u *Update) Check() error {
 		return fmt.Errorf("storage updates are not in order or unique")
 	}
 
+	// Find all created accounts by looking at all balance, nonce, and code updates
+	createdAccounts := make([]Address, 0, len(u.Balances)+len(u.Nonces)+len(u.Codes))
```
HerbertJordan: This might be misleading ... since what you are actually collecting here is all accounts with any value being updated, not only those accounts being created.

In the bigger picture, is this check still required? As I understand, `DeletedAccounts` is the next to be removed, and right now, our working theory is that `DeletedAccounts` is always empty. Maybe replace it wit ha check that `DeletedAccounts` is empty for now, and avoid the temporary complexity of trying to re-produce created accounts.

go/common/update_test.go
```diff
@@ -574,3 +508,37 @@ func TestUpdate_Print(t *testing.T) {
...
+	}
+
+	// Test inserting an address larger than all existing addresses
+	addr = Address{0x06}
+	updatedList = insertOrdered(list, addr)
+	expectedList = []Address{{0x01}, {0x03}, {0x05}, {0x06}}
+	if !reflect.DeepEqual(updatedList, expectedList) {
+		t.Errorf("expected %v, got %v", expectedList, updatedList)
+	}
+}
```
LorenzSchueler, line 527: nit: Consider making this a table test.

go/state/state_test.go
```diff
@@ -322,7 +324,8 @@ func TestCodeCanBeUpdated(t *testing.T) {
 	testEachConfiguration(t, func(t *testing.T, config *namedStateConfig, s state.State) {
 		// account must exist (not aut-created in Verkle Trie)
 		if _, err := s.Apply(0, common.Update{
-			CreatedAccounts: []common.Address{address1},
+			Balances: []common.BalanceUpdate{{Account: address1, Balance: balance1}},
+			Codes:    []common.CodeUpdate{{Account: address1, Code: []byte{}}},
```
LorenzSchueler, line 328: Why is the balance also set? Shouldn't the code be enough?

go/state/state_test.go
```diff
@@ -382,7 +385,8 @@ func TestCodeHashesMatchCodes(t *testing.T) {
...
```
LorenzSchueler, line 389: Why is the balance also set? Shouldn't the code be enough?

go/state/state_db_and_state_test.go
```diff
@@ -101,6 +105,7 @@ func TestCarmen_CanHandleMaximumBalance(t *testing.T) {
 
 			db.EndTransaction()
 			db.EndBlock(2)
+			require.NoError(t, db.Check(), "state DB check failed after block 1")
```
LorenzSchueler:
```suggestion
			require.NoError(t, db.Check(), "state DB check failed after block 2")
```

LorenzSchueler (dismissed):
Looks good.

go/state/state_test.go
ihtefyw: Not for the Rust implementation, which returns a default value if doesn't find the looked up value. And the default value of the code hash is the empty code hash
This is a discrepancy between the mpt and verkle implementation and there is no easy way to fix it.
However, this is not an issue when running the chain as nonce,balance, and code are always set together

go/state/state_test.go
ihtefyw: See above

go/common/update_test.go
```diff
@@ -574,3 +508,46 @@ func TestUpdate_Print(t *testing.T) {
...
 
 }
+
+func Test_insertOrdered(t *testing.T) {
+	cases := []struct {
+		name     string
+		list     []Address
+		addr     Address
+		expected []Address
+	}{
```
LorenzSchueler: nit: I've been told this is the idiomatic way to do it in Go
```suggestion
	cases := map[string]struct {
		list     []Address
		addr     Address
		expected []Address
	}{
```

LorenzSchueler (dismissed):

LorenzSchueler (dismissed):

HerbertJordan (dismissed):

LorenzSchueler (dismissed):

LorenzSchueler (dismissed):

LorenzSchueler (approved):

PR #422, opened 2026-05-26
Remove DeleteAccount from `State` implementations

ihtefyw:
This PR removes the `DeleteAccount` function from the `State` implementation.

Closes https://github.com/0xsoniclabs/sonic-admin/issues/762


go/database/mpt/root_hash_test.go
```diff
@@ -223,6 +223,8 @@ func TestS5RootHash_TwoAccountsWithExtensionNodeWithOddLength(t *testing.T) {
 	}
 }
 
+// TODO: Remove the hashes of the delete account at the bottom of the file
```
HerbertJordan: Is this a open TODO or should this line be removed?

go/state/state_db.go
```diff
@@ -1255,25 +1255,13 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
...
-				if s.accounts[addr].current != accountExists {
-					update.AppendDeleteAccount(addr)
-				}
-			} else {
+			if s.accounts[addr].original != accountExists {
 				nonExistingAccounts[addr] = true
 			}
-			// Increment the reincarnation counter of cleared addresses to invalidate
-			// cached entries in the stored data cache.
-			s.reincarnation[addr] = s.reincarnation[addr] + 1
```
HerbertJordan, line 1273: Follow-up: this may be the only place where the reincarnation map is updated. If so, it should be removable since all look-ups will just always return 0.

go/state/state_db_and_state_test.go
```diff
@@ -809,59 +809,6 @@ func TestStateDB_HasEmptyStorage_HandlesAccountSelfDestructCorrectly(t *testing.
...
-				statedb.AddBalance(address, balance)
-				statedb.EndTransaction()
-				statedb.EndBlock(blockNum)
-			}
-
-			// Case 1: deleted account because is empty
-			createAccountWithBalance(address1, balance1, 0)
-			statedb.BeginBlock()
-			statedb.BeginTransaction()
-			statedb.SubBalance(address1, balance1)
```
HerbertJordan, line 842: The implicit deletion of an account, by setting everything back to zero, might still be a thing. This part of the test should be preserved.

Example case:
- an externally owned account is initialized by a set-code transaction ([EIP-7702](https://eips.ethereum.org/EIPS/eip-7702)); its none and balance is zero, the code is a redirect;
- if the redirect is deleted, the code is set to 0, and now all 3 values of the account -- nonce, balance, and code are back to zero, and the account should no longer exist.

go/state/state_db_test.go
```diff
@@ -475,11 +474,6 @@ func TestStateDB_StoreDataCacheIsResetAfterSuicide(t *testing.T) {
 	// in the snapshot state nor is the account marked as being cleared. The value
 	// is retrieved from the store data cache.
 	db.BeginTransaction()
-	// This value is now fetched from the value store, which is supposed to be cleared
-	// at the end of the block deleting the account.
-	if got := db.GetState(address1, key1); got != zero {
```
HerbertJordan, line 478: It is unexpected that this check is removed -- in particular, since it fails if it is re-added. This may reveal an issue.

This test should be re-reviewed to clarify the expectations in the case of a "failed" suicide -- which may happen in practice.

go/database/mpt/diff.go
```diff
@@ -395,3 +384,69 @@ func recordValueUpdate(
...
+		Balance: &emptyBalance,
+		Nonce:   &emptyNonce,
+		Code:    &emptyCodeHash,
+	}
+	context.result[account.address] = diff
+	// Also record the deletion of all storage entries.
+	context.currentAccount = &account.address
+	return forEachStorageEntry(context.source, &account.storage, func(key common.Key, value common.Value) {
+		recordValueUpdate(context, key, common.Value{})
+	})
```
HerbertJordan: This should be the same as calling
```suggestion
	return collectDiff(context, triePosition{ref: account.storage}, triePosition{ref: emptyNodeReference})
```
which would re-use existing code. Does that work?

HerbertJordan (changes requested):

go/state/state_db.go
HerbertJordan: This mechanism might actually be needed after all: What if the same account is temporary created in a block and destroyed within it -- would the state from one block "bleed" over to the next?

go/database/mpt/root_hash_test.go
ihtefyw: I had to remove a couple of lines from the address and keys files

go/state/state_db.go
ihtefyw: I see your point. Without updating it, following blocks could query the deleted account state and get old values from the cache.
However I am thinking if this is now an overkill: an account can only be deleted once (in the same block), therefore the reincarnation value will either be 0 or 1. Therefore the reincarnation counters becomes like a set of past deleted accounts. Do you agree with me?

go/state/state_db_test.go
ihtefyw: I think this test tests something that can't be done anymore, e.g. suicide an account that has been created in a previous block.
Also the fact that this passes is wrong:
> 	db.EndBlock(2) // < here the stored data cache is reset to forget the old state; also, key2/val2 is stored in DB

This does not happen, an account is not created implicitly if a storage value is set. 
> 	if got := db.GetState(address1, key2); got != val2 {
>		t.Errorf("unexpected value, wanted %v, got %v", val2, got)
>	}

this passes only because the mock is set to return `val2` on `key2`

I think we can remove this.

ihtefyw: I removed it and substituted with a more fine-grained test
TestStateDB_QueryingStoredDataOfDestroyedAccountIsNotReturningDeletedValues

go/state/state_db.go
HerbertJordan: Is there a guarantee that the same smart contract address can not be re-used again?

The [CREATE2](https://www.evm.codes/#F5) instruction allows to create a contract at a controllable location. I think this could be used to re-create and self-destruct a contract in one block, and then do the same in the next block. The worry is that the state of the first block would remain visible in the next block.

go/state/state_db_test.go
```diff
@@ -422,68 +422,42 @@ func TestStateDB_StorageOfDestroyedAccountIsStillAccessibleTillEndOfTransaction(
...
-	if got := db.GetState(address1, key2); got != val2 {
-		t.Errorf("unexpected value, wanted %v, got %v", val2, got)
-	}
-	db.EndTransaction()
-	db.EndBlock(3)
+	// This will lookup in the storadDataCache, but since the reincarnation is higher, it's gonna return an empty value
+	got = db.GetState(address1, key1)
+	require.Equal(got, common.Value{})
+
 }
```
HerbertJordan, line 462: Can you add this test here?
```Go

func TestStateDB_StorageDataOfDestroyedAccountIsNotVisibleInNextTransactionOrBlock(t *testing.T) {
	ctrl := gomock.NewController(t)
	mock := NewMockState(ctrl)
	require := require.New(t)
	db := createStateDBWith(mock, 100, true)

	addr := common.Address{0x01}
	key := common.Key{0x01}
	val := common.Value{0x01}

	mock.EXPECT().GetBalance(addr).AnyTimes()
	mock.EXPECT().GetNonce(addr).AnyTimes()
	mock.EXPECT().GetCodeSize(addr).AnyTimes()
	mock.EXPECT().GetStorage(addr, key).AnyTimes()
	mock.EXPECT().Check().AnyTimes()
	mock.EXPECT().Apply(gomock.Any(), common.Update{}).AnyTimes()

	for range 5 {
		// Create a contract in a transaction, add data, and destroy the contract.
		db.BeginBlock()
		db.BeginTransaction()
		db.CreateContract(addr)
		require.Zero(db.GetState(addr, key))
		db.SetState(addr, key, val)
		require.Equal(val, db.GetState(addr, key))
		db.Suicide(addr)
		db.EndTransaction()

		// The modified state should be deleted.
		db.BeginTransaction()
		require.Zero(db.GetState(addr, key))
		db.EndTransaction()

		// The modified state should also be deleted in the next block.
		db.EndBlock(12)

		db.BeginBlock()
		db.BeginTransaction()
		require.Zero(db.GetState(addr, key))
		db.EndTransaction()
		db.EndBlock(14)
	}
}
```

HerbertJordan (approved):

PR #423, opened 2026-05-27
Fix errcheck lint errors

ihtefyw:
This PR fixes all the errcheck lints found by golangci-lint-v2.
In tests, they are treated with the `require` library.

This was partially automated

### DIFF
--- a/go/backend/stock/file/file.go
+++ b/go/backend/stock/file/file.go
@@ -184,7 +184,6 @@ func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V
 	if err != nil {
 		return meta, err
 	}
-	defer func() { err = errors.Join(err, freelist.Close()) }()
 	if err := verifyStackInternal[I](meta, freelist); err != nil {
 		return meta, err
 	}
--- a/go/backend/stock/file/file.go
+++ b/go/backend/stock/file/file.go
@@ -147,7 +147,7 @@ func verifyStockInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], di
 	return verifyStockFilesInternal[I](encoder, metafile, valuefile, freelistfile)
 }
 
-func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], metafile, valuefile, freelistfile string) (metadata, error) {
+func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], metafile, valuefile, freelistfile string) (m metadata, err error) {
 	// Attempt to parse the meta-data.
 	meta, err := utils.ReadJsonFile[metadata](metafile)
 	if err != nil {
@@ -184,7 +184,7 @@ func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V
 	if err != nil {
 		return meta, err
 	}
-	defer freelist.Close()
+	defer func() { err = errors.Join(err, freelist.Close()) }()
 	if err := verifyStackInternal[I](meta, freelist); err != nil {
 		return meta, err
 	}
@@ -211,7 +211,7 @@ func verifyStackInternal[I stock.Index](meta metadata, freelistfile utils.OsFile
 		if err != nil {
 			return err
 		}
-		defer stack.Close()
+		defer func() { err = errors.Join(err, stack.Close()) }()
 		list, err := stack.GetAll()
 		if err != nil {
 			return err
--- a/go/backend/stock/file/file.go
+++ b/go/backend/stock/file/file.go
@@ -147,7 +147,7 @@ func verifyStockInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], di
 	return verifyStockFilesInternal[I](encoder, metafile, valuefile, freelistfile)
 }
 
-func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], metafile, valuefile, freelistfile string) (m metadata, err error) {
+func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], metafile, valuefile, freelistfile string) (metadata, error) {
 	// Attempt to parse the meta-data.
 	meta, err := utils.ReadJsonFile[metadata](metafile)
 	if err != nil {
@@ -191,7 +191,7 @@ func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V
 	return meta, nil
 }
 
-func verifyStackInternal[I stock.Index](meta metadata, freelistfile utils.OsFile) error {
+func verifyStackInternal[I stock.Index](meta metadata, freelistfile utils.OsFile) (retErr error) {
 	// Check size of the free-list file.
 	{
 		indexSize := int(unsafe.Sizeof(I(0)))
@@ -210,7 +210,7 @@ func verifyStackInternal[I stock.Index](meta metadata, freelistfile utils.OsFile
 		if err != nil {
 			return err
 		}
-		defer func() { err = errors.Join(err, stack.Close()) }()
+		defer func() { retErr = errors.Join(retErr, stack.Close()) }()
 		list, err := stack.GetAll()
 		if err != nil {
 			return err
--- a/go/backend/stock/file/file_test.go
+++ b/go/backend/stock/file/file_test.go
@@ -1086,7 +1086,7 @@ func TestStock_Restore_FailsOnIoIssues(t *testing.T) {
 	}
 }
 
-func copyDirectory(src, dst string) (err error) {
+func copyDirectory(src, dst string) (retErr error) {
 	return filepath.Walk(src, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
@@ -1103,12 +1103,12 @@ func copyDirectory(src, dst string) (err error) {
 		if err != nil {
 			return err
 		}
-		defer func() { err = errors.Join(err, srcFile.Close()) }()
+		defer func() { retErr = errors.Join(retErr, srcFile.Close()) }()
 		dstFile, err := os.Create(dstPath)
 		if err != nil {
 			return err
 		}
-		defer func() { err = errors.Join(err, dstFile.Close()) }()
+		defer func() { retErr = errors.Join(retErr, dstFile.Close()) }()
 		_, err = io.Copy(dstFile, srcFile)
 		return err
 	})
--- a/go/backend/stock/file/file_test.go
+++ b/go/backend/stock/file/file_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/0xsoniclabs/carmen/go/backend/stock"
 	"github.com/0xsoniclabs/carmen/go/backend/utils"
 	"github.com/0xsoniclabs/carmen/go/backend/utils/checkpoint"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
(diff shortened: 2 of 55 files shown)


go/backend/stock/file/file.go
```diff
@@ -184,7 +184,7 @@ func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V
 	if err != nil {
 		return meta, err
 	}
-	defer freelist.Close()
+	defer func() { _ = freelist.Close() }()
```
HerbertJordan: In this context, it seems like the error should actually be handled. 
I think you can do so by naming the return value
```
func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V], metafile, valuefile, freelistfile string) (_ metadata, err error) {
```
and then write here

```suggestion
	defer func() { err = errors.Join(err, freelist.Close()) }()
```

go/backend/stock/file/file.go
```diff
@@ -211,7 +211,7 @@ func verifyStackInternal[I stock.Index](meta metadata, freelistfile utils.OsFile
 		if err != nil {
 			return err
 		}
-		defer stack.Close()
+		defer func() { _ = stack.Close() }()
```
HerbertJordan: same here

go/backend/stock/memory/memory.go
```diff
@@ -104,7 +104,7 @@ func OpenStock[I stock.Index, V any](encoder stock.ValueEncoder[V], directory st
 		if err != nil {
 			return nil, err
 		}
-		defer file.Close()
+		defer func() { _ = file.Close() }()
```
HerbertJordan: Also here, the error should not be ignored. If it is doable, those errors should be consumed.

go/backend/stock/memory/memory.go
```diff
@@ -248,11 +248,12 @@ func (s *inMemoryStock[I, V]) writeTo(dir string) error {
 	if f, err := os.Create(filepath.Join(dir, fileNameValues)); err != nil {
 		return err
 	} else {
-		defer f.Close()
+		defer func() { _ = f.Close() }()
 
 		buffer := make([]byte, s.encoder.GetEncodedSize())
 		for _, v := range s.values {
-			s.encoder.Store(buffer, &v)
+			// TODO: this could return the error instead of ignoring it
```
HerbertJordan: yes

go/backend/utils/buffered_file.go
```diff
@@ -52,12 +52,12 @@ func OpenBufferedFile(path string) (*BufferedFile, error) {
 func openBufferedFile(f OsFile) (*BufferedFile, error) {
 	stats, err := f.Stat()
 	if err != nil {
-		f.Close()
+		_ = f.Close()
 		return nil, err
```
HerbertJordan: How about

```suggestion
        return nil, errors.Join(err, f.Close())
```

go/carmen/block_test.go
```diff
@@ -332,7 +332,7 @@ func TestBlockContext_PanickingCommitsReleaseQueryLock(t *testing.T) {
 		}
 	}()
 
-	context.Commit()
+	_ = context.Commit()
```
HerbertJordan:
```suggestion
	require.NoError(t, context.Commit())
```

HerbertJordan (commented):
I went half through, and I think this PR should deal with all errors, unless there is an explicit reason for it. Right now, it looks like this PR is handling issues in tests where it is easy, by using `require.NoError`, but in the actual code it is hiding the issue by assigning it to `_`. This makes finding those issues just a bit harder in the future.

All those cases should actually be checked.

go/backend/stock/file/file.go
```diff
@@ -211,7 +211,7 @@ func verifyStackInternal[I stock.Index](meta metadata, freelistfile utils.OsFile
 		if err != nil {
 			return err
 		}
-		defer stack.Close()
+		defer func() { err = errors.Join(err, stack.Close()) }()
```
HerbertJordan: This error is only assigned to the local `err` variable, not the resulting error, since the error in the return value list is not named.

go/backend/stock/file/file_test.go
```diff
@@ -1102,12 +1103,12 @@ func copyDirectory(src, dst string) error {
 		if err != nil {
 			return err
 		}
-		defer srcFile.Close()
+		defer func() { err = errors.Join(err, srcFile.Close()) }()
```
HerbertJordan: This assigns the error to the parameter `err`, not to the resulting error, and thus it gets lost.

go/backend/stock/file/file_test.go
```diff
@@ -1102,12 +1103,12 @@ func copyDirectory(src, dst string) error {
...
 		dstFile, err := os.Create(dstPath)
 		if err != nil {
 			return err
 		}
-		defer dstFile.Close()
+		defer func() { err = errors.Join(err, dstFile.Close()) }()
```
HerbertJordan: Same here.

go/backend/stock/memory/memory.go
```diff
@@ -104,7 +105,7 @@ func OpenStock[I stock.Index, V any](encoder stock.ValueEncoder[V], directory st
 		if err != nil {
 			return nil, err
 		}
-		defer file.Close()
+		defer func() { err = errors.Join(err, file.Close()) }()
```
HerbertJordan: This `err` used here actually refers to the `err` declared in line 95, shadowing the named return error. So, the error assigned here is lost.

You can test this by
```Go
func createError() (int, error) {
	return 0, errors.New("an error occurred")
}

func myTest() (err error) {
	{ // < if you remove the scope, it works
		value, err := createError() // < creates a new error, shadowing the named return variable 'err'
		defer func() { err = errors.Join(err, fmt.Errorf("another error")) }()
		if err != nil {
			return err
		}

		_ = value // Use the value to avoid unused variable error
	}
	return nil
}

func TestErrorJoining(t *testing.T) {
	err := myTest()
	require.Error(t, err)
	require.ErrorContains(t, err, "an error occurred")
	require.ErrorContains(t, err, "another error")
}
```

HerbertJordan (changes requested):
I looked through the first few error handling corrections and spotted a number of issues due to variable name shadowing. Please have a look, and make sure the correct error variables are referenced.

go/backend/stock/memory/memory.go
HerbertJordan: A fix to avoid this is to rename the return-error to something different.

go/backend/stock/file/file.go
```diff
@@ -184,7 +184,6 @@ func verifyStockFilesInternal[I stock.Index, V any](encoder stock.ValueEncoder[V
 	if err != nil {
 		return meta, err
 	}
```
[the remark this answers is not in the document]
ihtefyw: It's close internally in `verifyStackInternal`

go/common/keccak.go
```diff
@@ -52,7 +60,12 @@ type keccakHasher interface {
 	Read(out []byte) (int, error)
 }
 
-var emptyKeccak256Hash = keccak256_Go([]byte{})
+var emptyKeccak256Hash Hash
+
+func init() {
+	// keccak256 of empty input cannot fail: the sha3 hasher always accepts Write/Read on valid buffers.
+	emptyKeccak256Hash, _ = keccak256_Go([]byte{})
```
HerbertJordan, line 67: nit: you can initialize it using keccak256_C.

go/database/mpt/tool/stress_test.go
```diff
@@ -88,7 +89,7 @@ func TestStressTest_ZeroBlocks(t *testing.T) {
 }
 
 func TestGetDirectorySize_NonExistentDirectory(t *testing.T) {
-	size := getDirectorySize("/path/does/not/exist")
+	size, _ := getDirectorySize("/path/does/not/exist")
```
HerbertJordan: I think the test should clarify whether the err should be nil or not in this case.

go/database/mpt/state_test.go
```diff
@@ -185,17 +185,17 @@ func BenchmarkStorageChanges(b *testing.B) {
 				}()
 
 				address := common.Address{}
-				state.SetNonce(address, common.ToNonce(12))
+				_ = state.SetNonce(address, common.ToNonce(12))
```
HerbertJordan: nit: The errors in this benchmark should be checked using a require.NoError(b, err), which should be straight forward.

go/state/tool/benchmark.go
```diff
@@ -277,7 +282,10 @@ func runBenchmarkState(
 
 			throughput := float64(reportingInterval*numInsertsPerBlock) / startReporting.Sub(lastReportTime).Seconds()
 			memory := state.GetMemoryFootprint().Total()
-			disk := getDirectorySize(path)
+			disk, err := getDirectorySize(path)
+			if err != nil {
```
HerbertJordan, line 284: nit: If an error occurred, the `disk` value should not be used below.

go/state/tool/benchmark.go
```diff
@@ -313,10 +321,10 @@ func runBenchmarkState(
 	return res, nil
 }
 
-// GetDirectorySize computes the size of all files in the given directory in bytes.
-func getDirectorySize(directory string) int64 {
+// getDirectorySize computes the size of all files in the given directory in bytes.
+func getDirectorySize(directory string) (int64, error) {
```
HerbertJordan: potential follow-up: looks like this function is duplicated ... it is defined in [here](https://github.com/0xsoniclabs/carmen/blob/f287070097dc5495198075140f628cc758dbf0e8/go/database/mpt/tool/stress.go#L390) and [here](https://github.com/0xsoniclabs/carmen/blob/f287070097dc5495198075140f628cc758dbf0e8/go/state/tool/benchmark.go#L325).

go/state/tool/benchmark_test.go
```diff
@@ -160,12 +160,13 @@ func TestGetDirectorySize_Normal(t *testing.T) {
...
 
-	size := getDirectorySize(dir)
+	size, err := getDirectorySize(dir)
+	require.NoError(t, err)
 	require.Equal(t, int64(len(data)), size)
 }
 
 func TestGetDirectorySize_NonExistentDir(t *testing.T) {
-	size := getDirectorySize("/path/does/not/exist")
+	size, _ := getDirectorySize("/path/does/not/exist")
```
HerbertJordan: It should be clarified whether in such a case an error is expected or not.

HerbertJordan (dismissed):
Looks good, a few minor fixes and some suggestions for follow-up tasks, then it is good to merge. Thanks!

go/common/keccak.go
ihtefyw: Actually I can't, keccack256_C refers to emptyKeccak256Hash, creating a cycle

go/database/mpt/state_test.go
ihtefyw: I thought to omit it for performance reason, as this is a benchmark
Will do

go/state/tool/benchmark.go
ihtefyw: That explains why I thought I already modified a couple of times...
Moved into a utils file and shared the implementation
Also the tests were duplicated

go/common/utils_test.go
```diff
@@ -0,0 +1,49 @@
...
+	size, _ := GetDirectorySize("/path/does/not/exist")
+	// GetDirectorySize should return 0 for non-existent directory
+	require.Equal(t, int64(0), size, "size should be zero for non-existent directory")
+}
+
+func TestGetDirectorySize_SkipsUnreadableFile(t *testing.T) {
+	dir := t.TempDir()
+	file := filepath.Join(dir, "file2")
+	data := []byte("xyz")
+	require.NoError(t, os.WriteFile(file, data, 0000)) // No permissions
```
HerbertJordan, line 43: nit: does this prevent the file from being cleaned up after the test?
If so, a defered chmod should be added.

HerbertJordan (approved):

go/common/utils_test.go
ihtefyw: I just tried it and no, it seems it still deletes the folder
I tried to remove write permissions on the test folder instead and interestingly the test fails with a `cleanup failed` error. So in go it's not possible to leak tmp folders

PR #424, opened 2026-05-27
Fix govet lints

ihtefyw:
This PR fixes all the govet lints found by golangci-lint-v2


LorenzSchueler (dismissed):
Why was this passing CI before? Is the linter not run at all?

LorenzSchueler (dismissed):

LorenzSchueler (approved):

PR #426, opened 2026-05-27
Fix staticcheck lints

ihtefyw:
This PR fixes the staticcheck lints, apart form the De Morgan laws one (which are going to be ignored)

### DIFF
--- a/go/.golangci.yml
+++ b/go/.golangci.yml
@@ -53,4 +53,8 @@ output:
       # for human consumption
       path: ./build/golangci-lint-report.html
     tab:
+<<<<<<< HEAD
       path: stdout
+=======
+      path: stdout
+>>>>>>> 164ef64a (Add .golangci.yml)
--- a/go/.golangci.yml
+++ b/go/.golangci.yml
@@ -53,8 +53,4 @@ output:
       # for human consumption
       path: ./build/golangci-lint-report.html
     tab:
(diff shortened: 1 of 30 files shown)


go/database/vt/reference/state_test.go
```diff
@@ -325,7 +326,7 @@ func TestState_Export_PanicsAsNotImplemented(t *testing.T) {
 	require := require.New(t)
 	state := newState()
 	require.Panics(
-		func() { state.Export(nil, nil) },
+		func() { state.Export(context.TODO(), nil) },
```
LorenzSchueler: I *THINK* `t.Context()` is the more suitable choice here. Please double-check.
```suggestion
		func() { state.Export(t.Context(), nil) },
```
Here and also in other files.

go/.golangci.yml
```diff
@@ -53,4 +53,8 @@ output:
       # for human consumption
       path: ./build/golangci-lint-report.html
     tab:
+<<<<<<< HEAD
       path: stdout
+=======
+      path: stdout
+>>>>>>> 164ef64a (Add .golangci.yml)
```
LorenzSchueler: git diff here 😆

LorenzSchueler (dismissed):

LorenzSchueler (approved):

PR #431, opened 2026-06-11
Remove `Exists` from `State` implementations

ihtefyw:
The existence of an account was tied to the its emptiness: every account theoretically exists, unless they are empty
Therefore, checking for the existence of an account corresponds to query it's values, e.g. balance, nonce, and code and check they are the default value

Having the `Exists` function in the `State` interface forced each implementation to implement the same logic, sometimes in different ways depending on the used data structure. Moreover, the existence of an account is a higher level concept, which belong in  the `StateDb` component.

This PR removes the `Exists` function from the `State` implementations and moves the emptiness check to the `StateDB::Exist` function. 
Importantly, this required to swap the creation of the account and the set of the nonce in `StateDB::setNonce`, as otherwise it would have incorrectly set the account as existing when setting the nonce for a non-existing account. This also aligns the `setNonce` behavior to the other set function, which creates the account first

To mark an account as already existing in tests, a mock expectation on `state.GetBalance` is used in most cases

### DIFF
--- a/go/state/state_db_test.go
+++ b/go/state/state_db_test.go
@@ -1612,7 +1612,6 @@ func TestStateDB_NoncesCanBeWrittenAndReadWithoutStateAccess(t *testing.T) {
 	mock := NewMockState(ctrl)
 	db := CreateStateDBUsing(mock)
 
-	// SetNonce creates the account if it does not exist
 	mock.EXPECT().GetBalance(address1).Return(balance1, nil)
 
 	var want uint64 = 12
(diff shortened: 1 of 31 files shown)


go/state/state_db_test.go
```diff
@@ -4986,3 +4947,49 @@ func TestStateDB_EndBlock_ClearsUndoList(t *testing.T) {
 	stateDB.EndBlock(12)
 	require.Empty(t, stateDB.undo)
 }
+
+func TestStateDB_Exist_ReturnsTrueForNonEmptyAccount(t *testing.T) {
```
HerbertJordan, line 4951: There should also be a test like this where the non-emptiness is not based on the temporary state in the state DB but from the persistent state in the state.

The practical difference: do not set the value through the StateDB to reach cached values but set it up in the state mock.

HerbertJordan (approved):

PR #433, opened 2026-07-13
Add `KVFile` and `KVCachedFile`

ihtefyw:
This PR adds the `KVFile` interface, which represents a set of operations to do on top of a Key-Value store, possibly backed by a file. 
The `KvCachedFile` decorates a `KVFile`, adding a caching layer and buffered writes.

Those two components will be used to implement caching for MPT codes and root list


go/backend/kv_file/kv_file.go
HerbertJordan, line 1: nit: Data structure interfaces like this are usually placed in the ./go/backend directory, not the common package. Should it be moved there?

go/backend/kv_file/kv_file.go
```diff
@@ -0,0 +1,57 @@
...
+type KVFileWithMemoryFootprint[K comparable, V any] interface {
+	KVFile[K, V]
+	GetMemoryFootprint() *MemoryFootprint
+}
+
+// readValueFn is a utility type for reading a key-value pair from an io.ReadSeeker.
+type readValueFn[K comparable, V any] func(reader io.ReadSeeker) (K, V, error)
+
+// writeValueFn is a utility type for writing a key-value pair to an io.Writer.
+type writeValueFn[K comparable, V any] func(writer io.Writer, key K, value V) error
```
HerbertJordan: nit: these are unused - should they be removed?

go/backend/kv_file/kv_file.go
```diff
@@ -0,0 +1,57 @@
...
+	Flush() error
+
+	// Size returns the number of key-value pairs in the file.
+	Size() (uint64, error)
+
+	// FileSize returns the size of the file in bytes.
+	FileSize() (uint64, error)
+
+	// GetAll retrieves all key-value pairs from the file.
+	GetAll() (map[K]V, error)
```
HerbertJordan: nit: if there is not a pressing use case for this, I would remove this from the interface, as the reason for having a file-backed key/value store is that the full map should not be in memory at the same time. If you only need it for tests and debugging in this package, maybe make it package private?

go/backend/kv_file/kv_cached_file.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	if dirty {
+		c.dirty[*key] = true
+	}
+	evictedKey, evictedValue, evicted := c.cache.Set(*key, *value)
+	if evicted {
+		if _, isDirty := c.dirty[evictedKey]; isDirty {
+			c.flushBuffer[evictedKey] = evictedValue
+			if len(c.flushBuffer) >= c.flushBufferThreshold {
+				return c.flushPending()
```
HerbertJordan: Not for this PR, but maybe considered as a follow-up: flush content in the background. The way this is implemented could lead to a long delay for a Get or Set operation whenever the buffer is full and needs to be evicted.

go/backend/kv_file/kv_cached_file.go
```diff
@@ -0,0 +1,284 @@
...
+	c.cache.Iterate(func(key K, value V) bool {
+		all[key] = value
+		return true
+	})
+
+	return all, nil
+}
+
+// Size returns the number of keys handled by the CachedFile.
+func (c *KVCachedFile[K, V]) Size() (uint64, error) {
```
HerbertJordan: There is a lot of complexity in this function just to support a `Size` operation. Is `Size()` really needed or could it be removed?

go/backend/kv_file/kv_cached_file.go
```diff
@@ -0,0 +1,284 @@
...
+// The cache keeps track of which cached entries are "dirty" (i.e. have not yet
+// been persisted to the underlying file). Only dirty entries are written on
+// flush, so repeated Flush calls with no intervening Set do not rewrite the
+// same values to disk.
+type KVCachedFile[K comparable, V any] struct {
+	cache       *LruCache[K, V]
+	flushBuffer map[K]V
+	dirty       map[K]bool
+	file        KVFileWithMemoryFootprint[K, V]
+	fromDisk    map[K]bool
```
HerbertJordan: `fromDisk` never shrinks, it just grows.

go/common/kv_cached_file.go
```diff
@@ -0,0 +1,284 @@
...
+		default:
+			return uintptr(unsafe.Sizeof(v))
+		}
+	}
+	mf := c.cache.GetDynamicMemoryFootprint(valueSize)
+	var sizeValues uintptr
+	for k, v := range c.flushBuffer {
+		sizeValues += unsafe.Sizeof(k) + valueSize(v)
+	}
+	return NewMemoryFootprint(unsafe.Sizeof(*c) + sizeValues + mf.Total() + fileFootprint.Total())
```
HerbertJordan: This function does not cover the dirty map nor the fromFile map. The nested file footprint should be registered as a sub-component, to forward its internal details.

HerbertJordan (changes requested):
It looks like the interface offers more than would be needed. Let me know if I am wrong about that, but I do not see the need for functions like `Size` and `GetAll` in the anticipated use cases (store codes and block-root hashes). If those get removed, the implementation could be simplified, since some of the internal fields become obsolete.

go/backend/kv_file/kv_file.go
ihtefyw: They will be used in the following PRs, sorry I should have added them later

go/backend/kv_file/kv_file.go
ihtefyw: Unfortunately, there are places for both the root list and codes where all the values are retrieved.
For example, in `VerifyMptState` ([verification.go](https://github.com/0xsoniclabs/carmen/blob/a916eb7b719650554bf61df8b7f567957e4c7726/go/database/mpt/verification.go#L60)) for the root list and in `GetCodes` ([state.go](https://github.com/0xsoniclabs/carmen/blob/a916eb7b719650554bf61df8b7f567957e4c7726/go/database/mpt/state.go#L357))

Because they are not used for the block processing and changing the implementations of the components that want all the data in memory touches a lot of code, I thought to leave a function returning the whole dataset for compatibility, as I don't want to change too many things in the release candidate.
After cherry-picking this commit to main, I want to then change this function to return an iterator or something similar.

What do you think?

go/backend/kv_file/kv_file.go
ihtefyw: mhhh, I was sure I got a circular import in the tests when I tried the first time, now it's gone..
I moved them

go/backend/kv_file/kv_file.go
ihtefyw: I see that you mentioned using an iterator in the codes PR, I will do that

go/common/kv_cached_file.go
ihtefyw: I didn't know about the sub-component, nice!

go/backend/kv_file/kv_cached_file.go
ihtefyw: It exists because I wanted a way to calculate the `Size()` in a lightweigth way
I may replace it by having a `Has()` method on the `KVFile`, which only checks if the key exists
or make fromDisk a cache too, putting a cap on it's size

go/backend/kv_file/kv_cached_file.go
ihtefyw: Looking at the changes in the other PRs, it could be removed entirely and replace by `FileSize()`, but it does make sense to be able to query how many elements are in file, what do you think?

go/backend/kv_file/kv_cached_file.go
ihtefyw: Opted for the Has() method

go/backend/kv_file/kv_cached_file.go
```diff
@@ -0,0 +1,269 @@
...
+// Entries found in the flushBuffer are promoted to the cache.
+func (c *KVCachedFile[K, V]) Get(key K) (*V, error) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	if val, inCache := c.cache.Get(key); inCache {
+		return &val, nil
+	}
+	if val, inFlushBuffer := c.flushBuffer[key]; inFlushBuffer {
+		delete(c.flushBuffer, key)
+		if err := c.handleCacheSet(&key, &val, false); err != nil {
```
HerbertJordan: Should the dirty flag be "true" instead of "false"? After all, you are fishing this value out from the flush buffer, which should represent a pending write.

go/backend/kv_file/kv_cached_file.go
```diff
@@ -0,0 +1,269 @@
...
+		default:
+			return uintptr(unsafe.Sizeof(v))
+		}
+	}
+	mf := c.cache.GetDynamicMemoryFootprint(valueSize)
+	var sizeValues uintptr
+	for k, v := range c.flushBuffer {
+		sizeValues += unsafe.Sizeof(k) + valueSize(v)
+	}
+	return common.NewMemoryFootprint(unsafe.Sizeof(*c) + sizeValues + mf.Total() + fileFootprint.Total())
```
HerbertJordan: The dirty map is not covered by this memory footprint report.

go/backend/kv_file/kv_cached_file.go
ihtefyw: Because it was in the flush buffer, it was already marked dirty in the first place, so it's not needed to mark it dirty again

ihtefyw: Double setting it to be sure :)

HerbertJordan (approved):

go/backend/kv_file/kv_cached_file.go
HerbertJordan: As long as it is not needed, I would not add it. A LevelDB instance is also a key-value store, but, as far as I know, it does not offer a simple way for getting the size because this is an IO expensive operation and given concurrent accesses, quite irrelevant information. Once you get a result, it is already obsolete.

PR #434, opened 2026-07-13
Cache codes using KVCachedFile

ihtefyw:
In the MPT implementation, all codes are kept in memory in a hash map. This increases memory consumption significantly and can possibly crash Carmen.
This PR adds a caching layer on top of the codes by using the `KVCachedFile`. Most of the existing code logic was generalized and moved to `OffsetFile`, a `KVFile` which stores the offset of each element in memory. 
The cache is set to 20,000 codes + 2,000 for the flush buffer, which equals to a worst case memory usage of ~1GB.

Notable changes:
- Because now getting a code could result in a file lookup, `GetCodeForHash` and `GetCodes` return an error now, and the interface and implementations have been updated accordingly.
- getReferencedCodes in `database/mpt/io/live.go` was still materializing the whole code file in memory. It has been modified to write the codes while scanning them, and renamed to `writeReferencedCodes`

Closes https://github.com/0xsoniclabs/sonic-admin/issues/844


go/database/mpt/codes.go
```diff
@@ -89,57 +76,36 @@ func openCodes(stateDirectory string) (*codes, error) {
...
-func (c *codes) getCodeForHash(hash common.Hash) []byte {
-	c.mutex.Lock()
-	res := c.codes[hash]
-	c.mutex.Unlock()
-	return res
+func (c *codes) getCodeForHash(hash common.Hash) ([]byte, error) {
+	code, err := c.codes.Get(hash)
+	if err != nil {
+		return []byte{}, err
+	}
```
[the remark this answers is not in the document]
ihtefyw: This was the default behavior already

go/common/offset_file.go
HerbertJordan, line 1: nit: make this contribution its own PR

go/common/offset_file.go
```diff
@@ -0,0 +1,256 @@
...
+		if err != nil {
+			return nil, err
+		}
+		res[key] = uint64(offset)
+	}
+	return res, nil
+}
+
+func readFromDiskAtOffset[K comparable, V any](path string, offset uint64, readValueFn readValueFn[K, V]) (*K, *V, error) {
+	file, err := os.Open(path)
```
HerbertJordan: Every read in case of a miss is opening and closing a file. Is this cheap? I would assume this to be rather costly. It should be possible keep the open file around for the life-time of the OffsetFile, to be able to reduce the Open, Seek, Close steps here to a single Seek.

go/database/mpt/codes.go
```diff
@@ -78,68 +71,50 @@ func openCodes(stateDirectory string) (*codes, error) {
...
+	return *code, nil
 }
 
 func (c *codes) getCodes() map[common.Hash][]byte {
-	c.mutex.Lock()
-	res := maps.Clone(c.codes)
-	c.mutex.Unlock()
-	return res
+	codes, err := c.codes.GetAll()
```
HerbertJordan: This function needs to be eliminated.

As far as I can see, the result of this call is only used for tests and for the export. Tests use small data sets, so it will never cause a problem. But the export may one day fail due an OOM because of this call.

Consider replacing it with some sort of iterator that you could return to walk through the codes instead of producing all at once.

go/common/offset_file.go
```diff
@@ -0,0 +1,256 @@
...
+		return err
+	}
+	defer func() {
+		err = errors.Join(err, file.Close())
+	}()
+
+	if err := o.writeEntryLocked(file, key, value); err != nil {
+		return err
+	}
+	return o.updateFileSizeLocked(file)
```
HerbertJordan: nit: avoid code duplication by calling SetBatch with one entry instead.

go/common/offset_file.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (o *OffsetFile[K, V]) SetBatch(pending map[K]V) (err error) {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+
+	file, err := os.OpenFile(o.filePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
```
HerbertJordan: Same as for read, every set goes through an Open, Seek, Close routine. You could keep the file open for the life-time of the OffsetFile.

HerbertJordan (changes requested):
Some initial feedback covering things that I think should be addressed before merging this. The changes may also be substantial enough that looking through the current tests in detail is premature, so I skipped that for now. Let me know what you think.

go/database/mpt/codes.go
```diff
@@ -59,87 +57,84 @@ func openCodes(stateDirectory string) (*codes, error) {
...
-	return res
+	// Preserve the historical non-nil contract so that callers can safely
+	// write into the returned map even after a read failure.
+	codes := map[common.Hash][]byte{}
+	seq, err := c.codes.Iterate()
+	if err != nil {
+		return codes
+	}
+	for key, code := range seq {
+		codes[key] = code
```
HerbertJordan: This is still building up a full code map. Instead of building the map here, the iterator should be returned.

go/database/mpt/codes.go
```diff
@@ -59,87 +57,84 @@ func openCodes(stateDirectory string) (*codes, error) {
...
 func (c *codes) GetMemoryFootprint() *common.MemoryFootprint {
-	var sizeCodes uint
-	c.mutex.Lock()
-	for k, v := range c.codes {
-		sizeCodes += uint(len(k) + len(v))
-	}
-	c.mutex.Unlock()
-	return common.NewMemoryFootprint(unsafe.Sizeof(*c) + uintptr(sizeCodes))
+	codes := c.codes.GetMemoryFootprint()
+	return common.NewMemoryFootprint(unsafe.Sizeof(*c) + codes.Total())
```
HerbertJordan: The memory footprint of `codes` should be reported as a child, to forward all its sub-structure.

go/database/mpt/codes.go
```diff
@@ -237,101 +236,83 @@ func (r codeRestorer) Restore(checkpoint checkpoint.Checkpoint) error {
 
 // readCodes parses the content of the given file if it exists or returns
 // a an empty code collection if there is no such file.
-func readCodes(path string) (map[common.Hash][]byte, error) {
-	codes, _, err := readCodesAndSize(path)
-	return codes, err
+func readCodes(path string) (codes map[common.Hash][]byte, err error) {
```
HerbertJordan: This function should return something the consumer can iterate over + close the file in the end, instead of the full code map. The code map may not fit into memory.

If it is only used for testing, rename it to `readCodesForTesting` to make indicate that this is not supposed to be used in production code.

go/database/mpt/codes.go
```diff
@@ -237,101 +236,83 @@ func (r codeRestorer) Restore(checkpoint checkpoint.Checkpoint) error {
...
 
-// readCodesAndSize parses the content of the given file and returns the
-// contained collection of codes and the size of the file.
-func readCodesAndSize(path string) (_ map[common.Hash][]byte, _ uint64, retErr error) {
-	// If there is no file, initialize and return an empty code collection.
-	info, err := os.Stat(path)
-	if os.IsNotExist(err) {
-		return map[common.Hash][]byte{}, 0, nil
+// writeCodes writes the given map of codes to the given file.
+func writeCodes(codes map[common.Hash][]byte, path string) (err error) {
```
HerbertJordan: I assume this is only used in tests => should be renamed like `readCodes` above.

go/database/mpt/codes.go
```diff
@@ -237,101 +236,83 @@ func (r codeRestorer) Restore(checkpoint checkpoint.Checkpoint) error {
...
+		err = errors.Join(err, storedCodes.Close())
+	}()
+	for hash, code := range codes {
+		err := storedCodes.Set(hash, code)
+		if err != nil {
+			return err
+		}
+	}
+	err = storedCodes.Flush()
+	return err
```
HerbertJordan:
```suggestion
	return errors.Join(
		storedCodes.SetBatch(codes),
		storedCodes.Flush(),
	)
```

HerbertJordan (approved):
Integration looks fine, but some functions still materialize the full code map. Those should be eliminated.

I did not check the unit tests.

PR #435, opened 2026-07-13
Cache RootLists with KVCachedFile

ihtefyw:
The `RootList` in the MPT archive implementation keeps all the root hashes in memory and can grow indefinitely.
Similarly to #434, this PR adds a caching layer to `RootList` by using the `KVCachedFile`. 
A new `KVFile` called `OrderedFile` is added, an array-like storage with a key-file offset mapping.

Closes https://github.com/0xsoniclabs/sonic-admin/issues/845

### DIFF
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -738,10 +738,11 @@ func loadRoots(archiveDirectory string) (_ *rootList, retErr error) {
 		)
 	}
 
-	size, err := kvFile.Size()
+	size, err := kvFile.FileSize()
 	if err != nil {
 		return nil, errors.Join(err, kvFile.Close())
 	}
+	size /= entrySize
 	if int(size) < checkpointData.NumRoots {
 		return nil, errors.Join(
 			fmt.Errorf("root list file is corrupted: expected at least %d roots, but found only %d", checkpointData.NumRoots, size),
--- a/go/database/mpt/archive_trie.go
+++ b/go/database/mpt/archive_trie.go
@@ -91,19 +91,19 @@ func OpenArchiveTrie(
 	forestConfig := ForestConfig{Mode: Immutable, NodeCacheConfig: cacheConfig}
 	forest, err := OpenFileForest(directory, config, forestConfig)
 	if err != nil {
-		return nil, err
+		return nil, errors.Join(err, roots.Close())
 	}
 
 	head, err := makeTrie(directory, forest)
 	if err != nil {
-		return nil, errors.Join(err, forest.Close())
+		return nil, errors.Join(err, roots.Close(), forest.Close())
 	}
 
 	root := NewNodeReference(EmptyId())
 	if roots.length() > 0 {
 		rootValue, err := roots.get(uint64(roots.length() - 1))
 		if err != nil {
-			return nil, err
+			return nil, errors.Join(err, roots.Close(), head.Close())
 		}
 		root = rootValue.NodeRef
 	}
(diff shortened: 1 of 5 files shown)


go/database/mpt/archive_trie.go
```diff
@@ -616,24 +650,53 @@ func (a *ArchiveTrie) GetConfig() MptConfig {
 // rootList is a utility type managing an in-memory copy of the list of roots
 // of an archive and its synchronization with an on-disk file copy.
 type rootList struct {
-	roots          []Root
-	filename       string // < the file storing the list of roots
-	directory      string // < the directory for checkpoint data
+	roots     *kv_file.KVCachedFile[uint64, Root] // < the in-memory copy of the roots list
```
HerbertJordan: nit: should use the interface instead of the concrete type

go/database/mpt/archive_trie.go
```diff
@@ -424,24 +459,17 @@ func (a *ArchiveTrie) GetMemoryFootprint() *common.MemoryFootprint {
 func (a *ArchiveTrie) Check() error {
 	roots := make([]*NodeReference, a.roots.length())
 	for i := 0; i < a.roots.length(); i++ {
-		roots[i] = &a.roots.roots[i].NodeRef
+		rootm, err := a.roots.get(uint64(i))
```
HerbertJordan, line 462: Here the full root list is loaded. The `Check()` function is mostly a debug tool for unit tests, but it is used by the (likely not well known) MPT tool. I quickly checked, the `CheckAll` function only uses the roots at a single place, where it could easily be replaced by an iterator. Maybe as a small follow-up PR.

go/database/mpt/archive_trie.go
```diff
@@ -644,107 +707,131 @@ func loadRoots(archiveDirectory string) (_ *rootList, retErr error) {
...
+	fileSize, err := kvFile.FileSize()
 	if err != nil {
-		return nil, err
+		return nil, errors.Join(err, kvFile.Close())
+	}
+	if fileSize%entrySize != 0 {
+		return nil, errors.Join(
+			fmt.Errorf("invalid root file format: size %d is not a multiple of entry size %d", fileSize, entrySize),
+			kvFile.Close(),
+		)
```
HerbertJordan: This check should not be of any concern on this level. This assumes internal implementation details of the KV file. If this needs to be checked, it should be inside the OpenOrderedFile factory.

go/database/mpt/archive_trie.go
```diff
@@ -644,107 +707,131 @@ func loadRoots(archiveDirectory string) (_ *rootList, retErr error) {
...
+// Iterate returns a sequence of all (block, root) pairs in the list.
+func (l *rootList) Iterate() (iter.Seq2[uint64, Root], error) {
+	return l.roots.Iterate()
 }
 
-func (l *rootList) storeRoots() error {
-	toBeWritten := l.roots[l.numRootsInFile:]
-	if l.numRootsInFile > 0 && len(toBeWritten) == 0 {
-		return nil
+func StoreRoots(filename string, rootsToWrite []Root) (err error) {
```
HerbertJordan, line 791: Also a follow-up: the parameter of this function is a full root list, which could be 100s of millions of entries. This should never materialize in memory.

go/database/mpt/archive_trie.go
HerbertJordan: This line needs to be updated to reflect the actual memory usage.

HerbertJordan (approved):

PR #437, opened 2026-07-21
Add Offset File

ihtefyw:
This PR adds the `OffsetFIle`, an append-only `KVFile` which keeps the offset of each element in memory.
Overwriting a value for a known key appends the pair at the end of the file and overwrites the offset of the previous key-value pair

This will be used to implement caching for the MPT codes

### DIFF
--- a/go/backend/kv_file/offset_file.go
+++ b/go/backend/kv_file/offset_file.go
@@ -1,0 +1,258 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package kv_file
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"iter"
+	"maps"
+	"os"
+	"sync"
+	"unsafe"
+
+	"github.com/0xsoniclabs/carmen/go/common"
+)
+
+// OffsetFile is a KVFile backed by a single append-only file with an in-memory
+// index mapping each key to the offset of its most recent record. Updating a
+// key appends a new record and repoints the index; existing records are never
+// modified.
+type OffsetFile[K comparable, V any] struct {
+	offsets map[K]uint64
+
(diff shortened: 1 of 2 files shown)


go/backend/kv_file/offset_file.go
```diff
@@ -0,0 +1,258 @@
...
+
+// writeEntryLocked appends a single (key, value) entry to the retained file
+// and updates the in-memory offset map. The caller must hold o.mutex and
+// must have verified that the file is not closed.
+func (o *OffsetFile[K, V]) writeEntryLocked(key K, value V) error {
+	curOffset, err := o.file.Seek(0, io.SeekEnd)
+	if err != nil {
+		return err
+	}
+	if err := o.writeValueFn(o.file, key, value); err != nil {
```
HerbertJordan, line 160: So, if I run a loop with
```
for range 100 {
    file.Set(12, 15)
}
```
the resulting file will contain 100 copies of the same pair, right?

Since you already have all keys cached, duplicates should be easily avoided. You know the position, and you could update the value in-place, instead of appending.

Related thought, feel free to ignore: If you want to force a no-update semantic, which would be sufficient for the Codes and root-hash use cases, you could update the KVFile interface to only have a `Get` and an `Add` function, where the latter fails if you try to add a new value with an existing key.

HerbertJordan (approved):

go/backend/kv_file/offset_file.go
ihtefyw: Thinking about it, I think it would be better to update the offset file to remove the append-only semantic: the KVFile should not restrict modification (also the offset file does not, it just forgets the past offset), and the non-updating logic can be easily implmemeted on top using the `Has()` method (that's how the codes does it)

There is indeed an issue with the record size, as you may overwrite with a bigger value that doesn't fit the space. In that case you may need to append the value and free the spot into a free list

ihtefyw: After trying, I think it's better to leave the file as it is: overwriting values comes with the issue that they may have variable size, and needs a logic to free space and append to the end in case the same key was updated with a bigger value. 
Moreover, you can reuse the freed spots only with new keys (otherwise the parsing logic of `last key wins` fails) and with exactly the same size (or the file would not be readable sequentially as you wouldn't know where the next record starts)

PR #438, opened 2026-07-21
Add Ordered file

ihtefyw:
This PR adds the `OrderedFile`, an array-like `KVFile`. 
Values have a fixed size, and their position in the file is determined by their key (uint64) times the size of the stored value.

This will be used to implement caching for the MPT root list

### DIFF
--- a/go/backend/kv_file/ordered_file.go
+++ b/go/backend/kv_file/ordered_file.go
@@ -43,6 +43,7 @@ type OrderedFile[V any] struct {
 
 // OpenOrderedFile opens an OrderedFile at the given path, creating the file if
 // it does not exist.
+// The key passed to `readValueFn` and `writeValueFn` is ignored, as the key is implicit in the file offset.
 func OpenOrderedFile[V any](path string, itemSize uint64, readValueFn readValueFn[uint64, V], writeValueFn writeValueFn[uint64, V]) (*OrderedFile[V], error) {
 	if _, err := os.Stat(path); os.IsNotExist(err) {
 		if err := os.WriteFile(path, []byte{}, 0600); err != nil {
@@ -110,15 +111,7 @@ func (o *OrderedFile[V]) Has(key uint64) (bool, error) {
 }
 
 func (o *OrderedFile[V]) Set(key uint64, value V) error {
-	o.mutex.Lock()
-	defer o.mutex.Unlock()
-
-	_, err := o.file.Seek(int64(key*o.itemSize), io.SeekStart)
-	if err != nil {
-		return err
-	}
-
-	return o.writeValueFn(o.file, key, value)
+	return o.SetBatch(map[uint64]V{key: value})
 }
 
 func (o *OrderedFile[V]) SetBatch(entries map[uint64]V) error {
--- a/go/backend/kv_file/ordered_file.go
+++ b/go/backend/kv_file/ordered_file.go
@@ -158,13 +158,6 @@ func (o *OrderedFile[V]) SetBatch(entries map[uint64]V) error {
 	return nil
 }
 
-func (o *OrderedFile[V]) Size() (uint64, error) {
-	o.mutex.Lock()
-	defer o.mutex.Unlock()
-
-	return o.sizeLocked()
-}
-
 func (o *OrderedFile[V]) FileSize() (uint64, error) {
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
--- a/go/backend/kv_file/ordered_file.go
+++ b/go/backend/kv_file/ordered_file.go
@@ -12,6 +12,7 @@ package kv_file
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"iter"
 	"os"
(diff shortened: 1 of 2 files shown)


go/backend/kv_file/ordered_file.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	return func(yield func(uint64, V) bool) {
+		for key := range size {
+			o.mutex.Lock()
+			value, err := o.readAtLocked(key)
+			o.mutex.Unlock()
+			if err != nil {
+				// Includes os.ErrClosed when the file is closed mid-iteration.
+				return
```
HerbertJordan, line 171: nit: those IO errors are lost; the caller would never know that it was not iterating through the full range.

Unfortunately, the current `Iterate()` signature does not really offer a good way out. It would have to be modified to something like
```
func (o *OrderedFile[V]) Iterate() (iter.Seq[Result[Entry[uint64,V]]], error) {
```
where `Result[T]` is a helper for a value or an error and `Entry[K,V]` is a key/value pair.

Not needed if you think this is an over-kill, just a thought I wanted to share.

go/backend/kv_file/ordered_file.go
```diff
@@ -0,0 +1,225 @@
...
+func (o *OrderedFile[V]) Set(key uint64, value V) error {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+
+	_, err := o.file.Seek(int64(key*o.itemSize), io.SeekStart)
+	if err != nil {
+		return err
+	}
+
+	return o.writeValueFn(o.file, key, value)
```
HerbertJordan: nit:
```suggestion
	return o.SetBatch(map[uint64]V{key: value})
```

go/backend/kv_file/ordered_file.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func (o *OrderedFile[V]) SetBatch(entries map[uint64]V) error {
```
HerbertJordan, line 117: Unlike a regular Key/Value file store, with sparse keys, a Set/SetBatch for a high number has a side-effect on all lower keys: it creates them and initializes them with zero.

For instance, if you call `Set(100, 42)` on an empty store, this implicitly creates all keys from 0-99 and gives them zero-values (in byte format, the interpretation is up to the reader). This is not the case for other KVFile implementations.

I guess for the intended use case this is not relevant, but I think it should be noted. Also, a test verifying the zero-initialization (and not a random initialization) may be useful.
```
func TestOrderedFile_Get_ReturnsZeroForNonInitializedKeys(t *testing.T) {
	require := require.New(t)
	f, _ := openTestOrderedFile(t)

	require.NoError(f.Set(1000, 42))

	for key := uint64(0); key < 1000; key++ {
		got, err := f.Get(key)
		require.NoError(err)
		require.NotNil(got)
		require.EqualValues(0, *got, "key %d", key)
	}
}
```

go/backend/kv_file/ordered_file.go
```diff
@@ -0,0 +1,225 @@
...
+	// mutex guards all file access. File access needs no explicit close-state
+	// tracking: os.File refcounts its descriptor internally, so any operation
+	// racing with (or following) Close fails with os.ErrClosed instead of
+	// touching a recycled descriptor.
+	mutex sync.Mutex
+}
+
+// OpenOrderedFile opens an OrderedFile at the given path, creating the file if
+// it does not exist.
+func OpenOrderedFile[V any](path string, itemSize uint64, readValueFn readValueFn[uint64, V], writeValueFn writeValueFn[uint64, V]) (*OrderedFile[V], error) {
```
HerbertJordan, line 47: I guess it should be "documented" that the key-part returned by the `readValueFn` is ignored and the `writeValueFn` should not bother writing the keys. Alternatively, a function definition only reading and writing values could be used. There is no pressing need to reuse the pre-existing types, since Open is not part of the interface.

go/backend/kv_file/ordered_file_test.go
```diff
@@ -0,0 +1,557 @@
...
+	next, _ := iter.Pull2(seq)
+	key, val, ok := next()
+	require.True(ok)
+	require.EqualValues(0, key)
+	require.EqualValues(1, val)
+
+	// Close the file before consuming the iterator.
+	require.NoError(f.Close())
+
+	_, _, ok = next()
```
HerbertJordan, line 371: A user of the file might be interested in knowing whether it got all the key/values or whether the iteration stopped before capturing all for one or the other reason.

HerbertJordan (approved):
I think the implementation fits the intended use case. However, I remarked a few design issues that you might want to address. Please have a look.

go/backend/kv_file/ordered_file.go
ihtefyw: Makes sense, but I don't think it belongs to this PR
I will add this to a subsequent one

go/backend/kv_file/ordered_file_test.go
ihtefyw: Because Iterate does not distinguish between closing the file and terminating the range, this is not easily possible. 
In a subsequent PR, I will change the Iterate signature to use a result which will allow to distinguish between the cases
I'm adding a TODO: to document it

PR #439, opened 2026-07-23
Make flush asynchronous in `kv_cached_file.go`

ihtefyw:
`KVCachedFile` currently hangs when the flushing threshold is reached, waiting for all the values to be written down to disk. This is expensive, especially if used for components active during block processing.
This PR makes the flush asynchronous, unblocking the user immediately after setting an element (conditionally).

The `maxPendingFlushes` threshold poses a limit to the memory pressure of the asynchronous updates, and adding a new element will wait until the number of elements to flush is lower than the upper bound.
It is set to `flushBufferThreshold + 1` in order to allow at least one completely asynchronous flush of the flush buffer

### DIFF
--- a/go/backend/kv_file/kv_cached_file.go
+++ b/go/backend/kv_file/kv_cached_file.go
@@ -31,6 +31,18 @@ type KVCachedFile[K comparable, V any] struct {
 	dirty       map[K]bool
 	file        KVFileWithMemoryFootprint[K, V]
 
+	// fileMu serializes every access to file, so the background writer and a
+	// foreground operation never touch the wrapped KVFile at the same time
+	// (KVFile is not required to be safe for concurrent use). It is independent
+	// of mu: the writer performs its disk I/O holding only fileMu, so foreground
+	// operations that stay in memory (cache/buffer hits, Set) proceed under mu
+	// meanwhile, and only those that must reach the file wait on fileMu.
+	//
+	// Lock order is mu then fileMu: a foreground caller may take fileMu while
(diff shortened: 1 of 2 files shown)


go/backend/kv_file/kv_cached_file.go
```diff
@@ -173,67 +258,163 @@ func (c *KVCachedFile[K, V]) GetMemoryFootprint() *common.MemoryFootprint {
...
 	var dirtySize uintptr
 	for k := range c.dirty {
 		dirtySize += unsafe.Sizeof(k) + unsafe.Sizeof(true)
 	}
-	return common.NewMemoryFootprint(unsafe.Sizeof(*c) + sizeValues + dirtySize + mf.Total() + fileFootprint.Total())
+	var pending uintptr
+	for _, buf := range c.pending {
+		pending += unsafe.Sizeof(buf)
+	}
+	return common.NewMemoryFootprint(unsafe.Sizeof(*c) + sizeValues + dirtySize + pending + mf.Total() + fileFootprint.Total())
```
HerbertJordan: nit: the sub-components like "fileFootprint", "c.pending" and "c.flushBuffer" could be listed as components of the memory footprint to provide details. See [AddChild](https://github.com/0xsoniclabs/carmen/blob/582fbf4cc9396a68146a2f1830712e79e8407f04/go/common/mem_footprint.go#L40).

HerbertJordan (approved):
The implementation looks good, but the tests are to complex for me to think through in reasonable time. As per your suggestion, I rely on you to have worked them out with proper care.

PR #455, opened 2026-08-04
Write cleared storage slots on account clearance

ihtefyw:
Clearing an account (re-creation via `CreateAccount` or the removal of empty or self-destructed accounts at the end of a transaction) used to reset the cached slot values of the account to `stored = 0, storedKnown= true`. This encoded the assumption that the clearing of the account physically purges its storage in the DB, making explicit zero-writes unnecessary.

That assumption no longer holds for one specific case. An account that is destroyed and stays destroyed is still purged completely: zeroing its balance, nonce, and code makes its account info empty, upon which the MPT removes the account node together with its entire storage. But if the account is re-created within the same block, its info is never empty at the end of the block, the node survives, and its former
storage is retained in the DB: nothing at the account level purges it, and slots whose final value matched the pretended zero stored value were skipped by the write-back in `EndBlock`.

This case cannot arise on post-Dencun chains: deploying onto an address with retained code or a non-zero nonce fails the create collision check (EIP-684), and an address holding storage without code can no longer come into existence, since SELFDESTRUCT either deletes nothing (for established contracts) or the entire account (for contracts created within the same transaction, whose storage never reaches the DB), as specified by EIP-6780. The only re-creation still occurring, deploying over an account holding nothing but a balance, involves no storage. Replaying pre-Dencun history (e.g. Sepolia), however, destroys and re-creates accounts with existing storage, which corrupted the state.

The clearing now leaves `stored` and `storedKnown` untouched, so they keep describing the value present in the DB. The `EndBlock` write-back then emits, for the cached slots of a cleared account:  
- an explicit zero for slots whose stored value is non-zero, purging the retained value;
- an explicit zero for slots whose stored value is unknown, making sure no former value is retained;
- an explicit value for slots re-written after the clearing, including writes of the zero value, which are now distinguishable from the cleared state;
- no write for slots whose final value equals their known stored value. For these, the stored data cache entry is refreshed with the account's incremented reincarnation counter: without the refresh, the still valid cached value would be masked as outdated and read as zero in future blocks, while the DB still holds the value.

The unit tests asserting the former contract are adapted:
- the clearing no longer resets `stored`/`storedKnown` of cached slots (`TestStateDB_CreateAccount_*`, `DeletingAccountWithData`);
- block updates now contain explicit zero-writes for cleared slots, including slots of accounts created and deleted within the same block, whose stored value is unknown (`RecreatingAnAccount*`, `RepeatedSuicide`, `ImplicitAccountCreatedBySetState*`, `SuicidedAccountNotRecreatedBySettingBalance`);
- the stored data cache is expected to hold the written zero with the current reincarnation instead of the masked stale value (`QueryingStoredDataOfDestroyedAccountIsNotReturningDeletedValues`).

**NOTE:** this fix does not address for stored values that are NOT cached. Those values will not be zeroed, and will persist in the underlying state. However, the scope of this PR is not to support pre-Dencun chains but rather to fix a behavior happening during the Sepolia history run.

Two new tests cover the explicit zero-write for a cleared slot with a known non-zero stored value, and the stored data cache refresh for a slot that is re-set to its stored value after a clearing.

Closes: https://github.com/0xsoniclabs/sonic-admin/issues/864

### DIFF
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -1294,19 +1294,16 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
 
 	// Update storage values in state DB
 	s.data.ForEach(func(slot slotId, value *slotValue) {
-		if value.storedKnown && value.stored == value.current {
-			// No DB write is needed, but if the account was cleared in this
-			// block its reincarnation counter was incremented; the cache
-			// entry is refreshed so that the still valid stored value is not
-			// masked as outdated by future reads. This covers slots that are
-			// explicitly set back to their stored value after a clearing.
-			if state, found := s.clearedAccounts[slot.addr]; found && (state == cleared || state == clearedAndTainted) {
-				s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
-			}
-			return
+		if !value.storedKnown || value.stored != value.current {
+			update.AppendSlotUpdate(slot.addr, slot.key, value.current)
+			s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
+		} else if state, found := s.clearedAccounts[slot.addr]; found && (state == cleared || state == clearedAndTainted) {
+			// The account was cleared in this block, incrementing its reincarnation
+			// counter; refresh the cache entry so the still valid stored value is
+			// not masked as outdated by future reads.
+			s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
 		}
-		update.AppendSlotUpdate(slot.addr, slot.key, value.current)
-		s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
+
 	})
 
 	// Update modified codes.
--- a/go/state/state_db.go
+++ b/go/state/state_db.go
@@ -544,9 +544,9 @@ func (s *stateDB) CreateAccount(addr common.Address) {
 				*value = backup
 			})
 
-			// Clear cached values.
-			value.stored = common.Value{}
-			value.storedKnown = true
+			// Clear cached values. The stored value and its known flag are
+			// deliberately left untouched: they keep describing the value
+			// currently present in the DB.
 			value.committed = common.Value{}
 			value.committedKnown = true
 			value.current = common.Value{}
@@ -1162,8 +1162,8 @@ func (s *stateDB) EndTransaction() {
 				if slot.addr == addr {
 					oldValue := *value
 					// Clear cached values.
-					value.stored = common.Value{}
-					value.storedKnown = true
+					// value.stored = common.Value{}
+					// value.storedKnown = true
 					value.committed = common.Value{}
 					value.committedKnown = true
 					value.current = common.Value{}
@@ -1296,10 +1296,19 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
 
 	// Update storage values in state DB
 	s.data.ForEach(func(slot slotId, value *slotValue) {
-		if !value.storedKnown || value.stored != value.current {
-			update.AppendSlotUpdate(slot.addr, slot.key, value.current)
-			s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
+		if value.storedKnown && value.stored == value.current {
+			// No DB write is needed, but if the account was cleared in this
+			// block its reincarnation counter was incremented; the cache
+			// entry is refreshed so that the still valid stored value is not
+			// masked as outdated by future reads. This covers slots that are
+			// explicitly set back to their stored value after a clearing.
+			if state, found := s.clearedAccounts[slot.addr]; found && (state == cleared || state == clearedAndTainted) {
+				s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
+			}
+			return
 		}
+		update.AppendSlotUpdate(slot.addr, slot.key, value.current)
+		s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
 	})
 
 	// Update modified codes.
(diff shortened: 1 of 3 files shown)


LorenzSchueler (commented):
This PR ensures that when the cache holds storage slots whose on-disk value is known to be non-zero for an account that gets deleted, the knowledge of the on-disk value is preserved (only the current value is cleared, via the new currentCleared flag) so that an explicit zero gets written to disk when the block ends.

This cannot happen in Sonic: starting with Cancun, self-destruct only deletes accounts created in the same transaction, whose storage never reached disk. The other clearing paths (redeployment over an existing address, empty-account deletion per EIP-161) are also unreachable, because on a chain with Cancun from genesis any account with on-disk storage has nonce ≥ 1 and code. However, chains that ran earlier forks can hit this, and it actually manifests in Sepolia at block 4803542.

This fix only resolves the issue for Sepolia. It suffices there because in all cases where the issue occurs, the affected slot is read during the block that deletes the account, so its non-zero on-disk value is in the cache. A general fix would also have to zero slots that are not cached. They are not just dead weight — they remain part of the state commitment, and if the address is later resurrected, the old storage can resurface as live state.

go/state/state_db.go
```diff
@@ -399,6 +399,12 @@ type slotValue struct {
 	storedKnown bool
 	// Whether the committed value is known.
 	committedKnown bool
+	// Marks the current value as cleared: the storage of the owning account
+	// was cleared during this block, and no explicit value has been set for
+	// this slot since. A cleared slot reads as zero, and at the end of the
+	// block a zero is written to the DB if the DB is known to hold a
+	// non-zero value for it.
+	currentCleared bool
```
LorenzSchueler, line 407: I think we could do without this flag. Just don't touch `stored` and `storedKnown` and set `current` to zero, as you already did. Then we always write 0, not only when we know the stored value, but I don't see any issue with that.

LorenzSchueler (changes requested):

go/state/state_db.go
ihtefyw: Not really: by not setting `stored` and `storedKnow=true`, when calling `EndBlock` in a tx where you create an acount, set some state, and then delete the account (okay for Dencun) you will have `storedKnow == false`, which will trigger a zero-write to the `State`.

This happens because, without the flag, the state `current = 0 & storedKnown = false` must be treated in the same way for both cleared and explicitly set values (you may want to explicitly set 0)

LorenzSchueler: And what is the problem with a zero write?

go/state/state_db.go
```diff
@@ -1296,10 +1296,19 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
 
 	// Update storage values in state DB
 	s.data.ForEach(func(slot slotId, value *slotValue) {
```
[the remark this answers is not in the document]
ihtefyw: This is intended. 
Values inside the MPT are not cleared

go/state/state_db.go
```diff
@@ -1296,10 +1294,19 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
...
+			// entry is refreshed so that the still valid stored value is not
+			// masked as outdated by future reads. This covers slots that are
+			// explicitly set back to their stored value after a clearing.
+			if state, found := s.clearedAccounts[slot.addr]; found && (state == cleared || state == clearedAndTainted) {
+				s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
+			}
+			return
 		}
+		update.AppendSlotUpdate(slot.addr, slot.key, value.current)
+		s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
```
LorenzSchueler: nit: This is the same behavior, but is looks better in the diff, making it obvious that only the reincarnation part was changed.
```suggestion
    if !value.storedKnown || value.stored != value.current {
        update.AppendSlotUpdate(slot.addr, slot.key, value.current)
        s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
    } else if state, found := s.clearedAccounts[slot.addr]; found && (state == cleared || state == clearedAndTainted) {
        // The account was cleared in this block, incrementing its reincarnation
        // counter; refresh the cache entry so the still valid stored value is
        // not masked as outdated by future reads.
        s.storedDataCache.Set(slot, storedDataCacheValue{value.current, s.reincarnation[slot.addr]})
    }
```

go/state/state_db_test.go
LorenzSchueler: Is this comment still true? Should it be updated as in in line 532?

LorenzSchueler (approved):
This PR ensures that when the cache holds storage slots of an account that gets cleared (deleted, or re-created over), the knowledge of the on-disk value is preserved: the clearing leaves stored/storedKnown untouched and only resets the current and committed values. At the end of the block, an explicit zero is then written to disk for every cached slot whose on-disk value is non-zero or unknown. Slots whose final value equals their known on-disk value get no write; their stored-data cache entry is instead refreshed with the incremented reincarnation counter, so the still-valid value is not masked as outdated by future reads.

This cannot happen in Sonic: starting with Cancun, self-destruct only deletes accounts created in the same transaction, whose storage never reached disk. The other clearing paths (redeployment over an existing address, empty-account deletion per EIP-161) are also unreachable, because on a chain with Cancun from genesis any account with on-disk storage has nonce ≥ 1 and code. However, chains that ran earlier forks can hit this, and it actually manifests in Sepolia at block 4803542.

This fix only resolves the issue for Sepolia. It suffices there because in all cases where the issue occurs, the affected slot is read before the deletion in the block that deletes the account, so its non-zero on-disk value is in the cache (reads after the clearing return zero without populating the cache). A general fix would also have to zero slots that are not cached. They are not just dead weight: they remain part of the state commitment, and since the deleting block already resurrected the account, they resurface as live state on the next read that misses the stored-data cache — the reincarnation counter only masks values already cached before the clearing.

go/state/state_db_test.go
ihtefyw: I updated the comment. 
With the new changes, it doesn't hit the lower reincarnation map shortcut but rather the cleared cached value directly, as the reincarnation counter has been updated before

LorenzSchueler (approved):

PR #464, opened 2026-08-06
Fuzz daily from this repository, rotating over its entry points

LuisPH3:
Moves Carmen's fuzzing out of LaScala and into this repository, where it is easier to maintain: the pipeline sits next to the code it fuzzes, and builds on the same `CI/Dockerfile.jenkins` the PR job already uses.

That is also why it has to live here. A Jenkins `dockerfile` agent is resolved against the *pipeline's own* SCM, so a pipeline in LaScala cannot use this image — it built Carmen with whatever the agent happened to have installed instead, none of it pinned to what this repository expects. Tosca's fuzzer broke exactly that way when its Rust version moved into the image; Carmen has the widest toolchain of the three and was next in line.

`CI/fuzzing.jenkinsfile` runs once a day and fuzzes one entry point, rotating over every `func FuzzXxx(f *testing.F)` it finds in its own checkout — so adding a fuzzer needs no change to this pipeline. `EntryPoint` is for running one by hand; empty means whichever has gone longest without a turn. The rotation and the fuzzing steps come from LaScala's shared library (0xsoniclabs/lascala#312, merged).

One line worth not deleting: `GOCACHE` points into the workspace because `go test -fuzz` keeps its corpus in `$GOCACHE/fuzz`, and the container does not survive the build.

Note the current entry points are all pure Go, so `go test` runs without the `carmen_cpp`/`carmen_rust` tags the PR job passes — same as before, but a fuzzer needing the native bindings would need them.

Needs a Jenkins job pointed at this file; nothing runs until that exists.

### DIFF
--- a/CI/fuzzing.jenkinsfile
+++ b/CI/fuzzing.jenkinsfile
@@ -1,0 +1,159 @@
+// Copyright (c) 2025 Sonic Operations Ltd
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at soniclabs.com/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+// Fuzzes one Carmen entry point, built on the toolchain image this commit ships
+// (CI/Dockerfile.jenkins), so what is fuzzed and what it is built with can never drift apart. That
+// is why this pipeline lives here rather than in LaScala: a Jenkins dockerfile agent is resolved
+// against the pipeline's own repository, so only a pipeline in Carmen can use Carmen's image.
+//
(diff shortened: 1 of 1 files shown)


CI/fuzzing.jenkinsfile
```diff
@@ -0,0 +1,134 @@
...
+        buildDiscarder(logRotator(numToKeepStr: '200'))
+    }
+
+    triggers {
+        cron('@daily')
+    }
+
+    environment {
+        CC = 'clang-19'
+        CXX = 'clang++-19'
```
ihtefyw, line 48: Those are not required anymore in main

ihtefyw (approved):

CI/fuzzing.jenkinsfile
LuisPH3: follow up

PR #472, opened 2026-08-07
Use root iterator in Check* functions

ihtefyw:
The `State::CheckAll` function requires an array with all the root node references. 
This required getting an iterator from the `rootList` and consume it to construct the array.

This PR fixes this by working on the iterator directly, which is the new standard way of working with the root list


go/database/mpt/archive_trie.go
```diff
@@ -457,17 +458,16 @@ func (a *ArchiveTrie) GetMemoryFootprint() *common.MemoryFootprint {
...
 
 func (a *ArchiveTrie) Check() error {
-	roots := make([]*NodeReference, a.roots.length())
-	for i := 0; i < a.roots.length(); i++ {
-		rootm, err := a.roots.get(uint64(i))
-		if err != nil {
-			return err
-		}
-		roots[i] = &rootm.NodeRef
+	it, err := a.roots.Iterate()
```
LorenzSchueler, line 464: This will silently pass if a read in `OrderedFile.Iterate` fails, because the iterator just stops.

go/database/mpt/nodes.go
LorenzSchueler, line 500: Here we still materialize all root node references, so reading lazily from an iterator does not buy us anything.

go/common/iter_utils/iter.go
```diff
@@ -0,0 +1,40 @@
...
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package iter_utils
+
+import "iter"
+
+// Map2 maps an `iter.Seq2` of key-value pairs to another `iter.Seq2`
+// of key-value pairs using the provided mapping function.
+func Map2[KIn, VIn, KOut, VOut any](
```
LorenzSchueler: The key is never consumed. A `Map` would be sufficient.

go/common/iter_utils/iter.go
```diff
@@ -0,0 +1,40 @@
...
+			if !yield(f(k, v)) {
+				return
+			}
+		}
+	}
+}
+
+// FromSliceWith creates an `iter.Seq2` from a slice of values,
+// where the keys are generated by the provided key function.
+func FromSliceWith[K comparable, V any](slice []V, keyFunc func(V) K) iter.Seq2[K, V] {
```
LorenzSchueler: The key is never consumed. Using `iter.Seq` and removing the `keyFunc` altogether would remove a majority of the diff.

go/database/mpt/archive_trie.go
LorenzSchueler, line 464: Do we need to acquire `rootsMutex` here?

go/common/iter_utils/iter_test.go
```diff
@@ -0,0 +1,84 @@
...
+	})
+
+	// Define the expected results after mapping
+	expectedResults := map[string]int{
+		"one":   10,
+		"two":   20,
+		"three": 30,
+	}
+
+	require.Equal(expectedResults, results)
```
LorenzSchueler:
```suggestion
	input := map[int]string{1: "one", 2: "two", 3: "three"}
	expected := map[string]int{"one": 10, "two": 20, "three": 30}

	mapFunc := func(k int, v string) (string, int) {
		return v, k * 10
	}

	seq := Map2(maps.All(input), mapFunc)
	require.Equal(t, expected, maps.Collect(seq))
```
WDYT? Similarly below.

LorenzSchueler (changes requested):
At least in the current state, I don't see what this PR is buying us. It looks to me like the full iterator gets materialized. Also, the keys are never read, only the values. Using `Map` and `Seq` would cut a lot of noise.

go/database/mpt/archive_trie.go
ihtefyw: The issue with Iterator ignoring errors will be addressed in a following PR, it requires quite a lot of changes

ihtefyw: I suppose we want to lock `rootsMutex`, but not doing it would not lead to data races as `KVCachedFile` is thread safe, and `Iterate` locks the file before reading from it.
But probably we want to lock it to avoid other components to push more blocks into the archive

go/database/mpt/nodes.go
ihtefyw: But the interface does not force that anymore.
The fact that the algorithm still materializes the iterator is unfortunate, and will be addressed in another PR

go/common/iter_utils/iter.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+		}
+	}
+}
+
+// FromSlice creates an `iter.Seq` from a slice of values.
+// Original ordering is preserved.
+func FromSlice[V any](slice []V) iter.Seq[V] {
+	return DropKeys(slices.All(slice))
+}
```
LorenzSchueler:
```suggestion
```

You can just use `slices.Values` at the call sites instead.

go/database/mpt/forest.go
```diff
@@ -673,12 +675,12 @@ func (s *Forest) Dump(rootRef *NodeReference) {
 // errors are detected, the Trie is to be considered in an invalid state and
 // the behavior of all other operations is undefined.
 func (s *Forest) Check(rootRef *NodeReference) error {
-	return s.CheckAll([]*NodeReference{rootRef})
+	return s.CheckAll(iter_utils.FromSlice([]*NodeReference{rootRef}))
```
LorenzSchueler: here and in other call sites:
```suggestion
	return s.CheckAll(slices.Values([]*NodeReference{rootRef}))
```

LorenzSchueler (approved):

PR #475, opened 2026-08-10
Add `_mocks.go` to skip list

ihtefyw:
The license header script was erroneously adding the license to mock files with multiple mocks inside which werern't starting with the line `// Code generated` (e.g. `external_state_mocks.go`)
This PR fixes it


LorenzSchueler (approved):
Maybe also take a look at this script in sonic. It is slightly different, and we should maybe try to keep them in sync.

PR #477, opened 2026-08-11
Add `ResultSeq` to iter_utils

ihtefyw:
This PR adds a couple of utility to the `iter_utils` module, in preparation for returning errors from the `KVFile::Iterate` function.

It adds the new type `ResultSeq[V]`, which is an iterator over a `Result[V]`, and a couple of ergonomic functions.
The key-value counterpart is `ResultSeq2[K,V]`, which is defined as an iterator over `Result[Pair[K,V]]`. This was preferred instead of using a `iter.Seq2` to allow both key and value into the result.

To iterate it, the user calls `Unwrap[2]`, which returns the underlying value iterator and a function to query for possible errors.
When the iterator stops, the user must call the error functions to check the iterator ending status

### DIFF
--- a/go/common/iter_utils/iter.go
+++ b/go/common/iter_utils/iter.go
@@ -12,6 +12,7 @@ package iter_utils
 
 import (
 	"iter"
+	"slices"
 
 	"github.com/0xsoniclabs/carmen/go/common/result"
 )
@@ -39,22 +40,29 @@ type ResultSeq[V any] = iter.Seq[result.Result[V]]
 // equivalent of an `iter.Seq2`.
 type ResultSeq2[K, V any] = ResultSeq[Pair[K, V]]
 
-// OkSeq lifts an `iter.Seq` into a `ResultSeq` that never fails.
-func OkSeq[V any](seq iter.Seq[V]) ResultSeq[V] {
-	return func(yield func(result.Result[V]) bool) {
+// Map maps an `iter.Seq` of values to another `iter.Seq` of values using the provided mapping function.
+func Map[VIn, VOut any](
+	seq iter.Seq[VIn],
+	f func(VIn) VOut,
+) iter.Seq[VOut] {
+	return func(yield func(VOut) bool) {
 		for v := range seq {
-			if !yield(result.Ok(v)) {
+			if !yield(f(v)) {
 				return
 			}
 		}
 	}
 }
 
-// OkSeq2 lifts an `iter.Seq2` into a `ResultSeq2` that never fails.
-func OkSeq2[K, V any](seq iter.Seq2[K, V]) ResultSeq2[K, V] {
-	return func(yield func(result.Result[Pair[K, V]]) bool) {
+// Map2 maps an `iter.Seq2` of key-value pairs to another `iter.Seq2` of key-value pairs using the provided mapping function.
+func Map2[KIn, VIn, KOut, VOut any](
+	seq iter.Seq2[KIn, VIn],
+	f func(KIn, VIn) (KOut, VOut),
+) iter.Seq2[KOut, VOut] {
+	return func(yield func(KOut, VOut) bool) {
 		for k, v := range seq {
-			if !yield(result.Ok(Pair[K, V]{Key: k, Value: v})) {
+			k2, v2 := f(k, v)
+			if !yield(k2, v2) {
 				return
 			}
 		}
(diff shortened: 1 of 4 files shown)


go/common/iter_utils/iter.go
```diff
@@ -12,8 +12,87 @@ package iter_utils
...
 )
 
+// Pair combines a key and a value into a single value. It allows key-value
+// pairs to be transported through the single-element `iter.Seq`, which is
+// required whenever the elements need to be wrapped, for instance in a
+// `result.Result`.
+type Pair[K, V any] struct {
+	Key   K
+	Value V
+}
```
LorenzSchueler, line 27: If Go only had proper tuple types 🙄

go/common/iter_utils/iter.go
```diff
@@ -12,8 +12,87 @@ package iter_utils
...
+
+// OkSeq lifts an `iter.Seq` into a `ResultSeq` that never fails.
+func OkSeq[V any](seq iter.Seq[V]) ResultSeq[V] {
+	return func(yield func(result.Result[V]) bool) {
+		for v := range seq {
+			if !yield(result.Ok(v)) {
+				return
+			}
+		}
+	}
```
LorenzSchueler:
```suggestion
	return Map(seq, result.Ok[V])
```
But at this point we don't even need this helper.

go/common/iter_utils/iter.go
```diff
@@ -12,8 +12,87 @@ package iter_utils
...
+}
+
+// OkSeq2 lifts an `iter.Seq2` into a `ResultSeq2` that never fails.
+func OkSeq2[K, V any](seq iter.Seq2[K, V]) ResultSeq2[K, V] {
```
LorenzSchueler: Similar to above. Consider removing `OkSeq2` and adding a `Map2` which is more flexible.

go/common/iter_utils/iter.go
```diff
@@ -28,7 +107,37 @@ func Map[VIn, VOut any](
...
+func MapOk[VIn, VOut any](
+	seq ResultSeq[VIn],
+	f func(VIn) VOut,
+) ResultSeq[VOut] {
+	return Map(seq, func(r result.Result[VIn]) result.Result[VOut] {
+		v, err := r.Get()
+		if err != nil {
+			return result.Err[VOut](err)
+		}
+		return result.Ok(f(v))
```
LorenzSchueler: Please try to build tools which integrate well with each other and are more generic and usable in many contexts.

```suggestion
		return result.Map(r, f)
```
and in result.go
```go
// Map applies the given function to the value of a successful Result. A failed
// Result is forwarded unchanged.
func Map[TIn, TOut any](r Result[TIn], f func(TIn) TOut) Result[TOut] {
	value, err := r.Get()
	if err != nil {
		return Err[TOut](err)
	}
	return Ok(f(value))
}
```

go/common/iter_utils/iter.go
```diff
@@ -28,7 +107,37 @@ func Map[VIn, VOut any](
...
+	f func(KIn, VIn) (KOut, VOut),
+) ResultSeq2[KOut, VOut] {
+	return Map(seq, func(r result.Result[Pair[KIn, VIn]]) result.Result[Pair[KOut, VOut]] {
+		in, err := r.Get()
+		if err != nil {
+			return result.Err[Pair[KOut, VOut]](err)
+		}
+		key, value := f(in.Unpack())
+		return result.Ok(Pair[KOut, VOut]{Key: key, Value: value})
+	})
```
LorenzSchueler:
```suggestion
	return MapOk(seq, func(in Pair[KIn, VIn]) Pair[KOut, VOut] {
		key, value := f(in.Unpack())
		return Pair[KOut, VOut]{Key: key, Value: value}
	})
```

go/common/iter_utils/iter.go
```diff
@@ -38,3 +147,21 @@ func DropKeys[K, V any](seq iter.Seq2[K, V]) iter.Seq[V] {
...
+// Enumerate creates an `iter.Seq2` from a slice of values, pairing each value
+// with its `uint64` index.
+func Enumerate[V any](slice []V) iter.Seq2[uint64, V] {
+	return func(yield func(uint64, V) bool) {
+		for i, v := range slice {
+			if !yield(uint64(i), v) {
+				return
+			}
+		}
+	}
```
LorenzSchueler: When we have `Map2` this could become something like `return Map2(slices.All(slice), func(k, v) { uint64(k), v })`

go/common/iter_utils/iter_test.go
```diff
@@ -11,14 +11,92 @@
...
+			return
+		}
+		yield(result.Ok("three"))
+	})
+
+	require.Equal([]string{"one"}, slices.Collect(seq))
+	require.ErrorIs(seqErr(), injected)
+}
+
+func TestIter_Unwrap2_YieldsAllPairsWithoutError(t *testing.T) {
```
LorenzSchueler, line 80: Unwrap counterpart missing.

go/common/iter_utils/iter.go
ihtefyw: For OkSeq2 is more complicated, as it converts an iter.Seq2 into an iter.Seq
For completeness I would keep also OkSeq

go/common/iter_utils/iter_test.go
```diff
@@ -30,12 +117,96 @@ func TestIter_Map_ReturnsMappedSeq(t *testing.T) {
 	require.Equal(t, expected, slices.Collect(seq))
 }
```
LorenzSchueler, line 118: Map2 counterpart missing.

LorenzSchueler: Also, please take a look at the suppressed comments by copilot above. They seem reasonable.

ihtefyw: It's nice how it suppresses interesting comments and shouts about imaginary build failures....

LorenzSchueler (approved):

PR #478, opened 2026-08-11
Return Result from `KVFile::Iterate`

ihtefyw:
Currently, `KVFile::Iterate` returns only values. Errors causes the iterator to silently stop without notifying the user.

This PR fixes it by making it return a `ResultSeq2`, n iterator over a `Result[Pair[K,V]]`, which is a type that can hold either a value or an error.
The `Database::CheckAll` interface function signature has been updated to reflect the change, and now takes an `ResultSeq`.

Closes https://github.com/0xsoniclabs/sonic-admin/issues/881

### DIFF
--- a/go/database/mpt/codes.go
+++ b/go/database/mpt/codes.go
@@ -16,8 +16,6 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"iter"
-	"maps"
 	"os"
 	"path/filepath"
 	"unsafe"
@@ -26,6 +24,7 @@ import (
 	"github.com/0xsoniclabs/carmen/go/backend/utils"
 	"github.com/0xsoniclabs/carmen/go/backend/utils/checkpoint"
 	"github.com/0xsoniclabs/carmen/go/common"
+	"github.com/0xsoniclabs/carmen/go/common/iter_utils"
 	"golang.org/x/crypto/sha3"
 )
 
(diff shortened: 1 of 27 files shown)


go/database/mpt/codes_test.go
```diff
@@ -342,9 +345,12 @@ func TestCodes_getCodes_IteratorTerminatesGracefullyOnClose(t *testing.T) {
 
 	// After the store has been closed, iterating the previously obtained
 	// sequence must not panic; it may yield fewer entries than the store
-	// contained but must terminate gracefully.
-	for range it {
+	// contained, but the close must be reported instead of silently
+	// truncating the iteration.
+	pairs, itErr := iter_utils.Unwrap2(it)
+	for range pairs {
 	}
```
LorenzSchueler: We have to ensure that the iterator is consumed before checking the error. Can we be sure this loop is not optimized away?

go/database/mpt/codes.go
```diff
@@ -122,7 +122,12 @@ func (c *codes) getCodesForTesting() (map[common.Hash][]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return maps.Collect(it), nil
+	seq, seqErr := iter_utils.Unwrap2(it)
+	res := maps.Collect(seq)
+	if err := seqErr(); err != nil {
+		return nil, err
+	}
+	return res, nil
```
LorenzSchueler: Didn't you have a helper for that somewhere?

ihtefyw: I had this patter a bit everywhere, I added a `CollectOk[2]` family of functions

go/database/mpt/codes_test.go
ihtefyw: Changed to use `CollectOk2`

LorenzSchueler (approved):

PR #479, opened 2026-08-11
Substitute experimental packages with standard library ones

ihtefyw:
Some of the packages used in Carmen were originally only available in the golang experimental libraries.
This PR substituite some of them with the standard library ones, updating the functions accordingly.

In particular:
- `maps.Keys` returns now an iterator instead of a slice
- `slices.SortFunc` wants the comparison function to return an int and not a boolena


LuisPH3 (approved):
nice

LorenzSchueler (approved):
👍🏽

PR #480, opened 2026-08-11
Modernize code with `go fix`

ihtefyw:
This PR modernizes the code by using the `go fix` command.

This was run with `go 1.27rc`, as the one included in `go 1.26.3` includes a bug which does not declare variables after substituing a map loop copy with `maps.Collect`


LuisPH3 (approved):
I looks like interfaces are not modified.
Mostly min-max and ranges.

sort namespace changed for slices.sort, it seems that they aliased since go 1.22, no change there.

PR #484, opened 2026-09-01
Refactor `InterTxSnapshot` utility functions

ihtefyw:
This PR moves some utility functions and types used by the `InterTxSnapshot` combinatorial test to a separate test utils file. 
Those will be used to implement combinatorial testing for the block rollback mechanism in a following PR.

Additional changes:
- `OpArgs` now uses values instead of pointers to simplify the interface


LorenzSchueler (commented):
What happened? Not code is not moved but deleted.

LorenzSchueler (approved):

PR #485, opened 2026-09-07
Remove allocation from `LruCache::Clear`

ihtefyw:
The `LruCache::Clear` implementation was clearing the cache by allocating a new map each time. This is critical for performance when `Clear` is frequently called.

The new implementation just clear the map instead.


LuisPH3 (approved):
👍

LorenzSchueler (approved):

PR #486, opened 2026-09-07
Fix panic on single element `LruCache` on `dropLast`

ihtefyw:
`LruCache::dropLast` panics when removing the last element of a single-element cache.

This PR fixes it, and adds a test to detect the behavior


go/common/lru_cache_test.go
```diff
@@ -125,3 +125,28 @@ func TestCache_Entry_String(t *testing.T) {
...
+func TestLruCache_dropLast_DoesNotPanicOnEmptyOrSingleValueCache(t *testing.T) {
+	testCases := map[string]func() *LruCache[int, int]{
+		"empty": func() *LruCache[int, int] {
+			return NewLruCache[int, int](4)
+		},
+		"single value": func() *LruCache[int, int] {
+			cache := NewLruCache[int, int](4)
+			cache.Set(1, 2)
+			return cache
+		},
```
LorenzSchueler, line 180: Please also add a test that it still works with multiple elements.

go/common/lru_cache.go
```diff
@@ -182,10 +182,17 @@ func (c *LruCache[K, V]) touch(item *entry[K, V]) {
 
 // dropLast drop the last element from the queue and returns it
 func (c *LruCache[K, V]) dropLast() (dropped *entry[K, V]) {
```
LorenzSchueler, line 184: The fix itself is fine but IMO the function should not hand out an entry which still has `prev` and `next` pointers, especially because `prev` still points into the cache. Returning `(K, V)` would be much cleaner. Not sure if you want to include this change here though.

ihtefyw: I agree, but I can do this in another pr

go/common/lru_cache_test.go
ihtefyw: Added an additional test

go/common/lru_cache.go
ihtefyw: I decided to do it here because writing the tests with the `entry` structure was quite annoying.

LorenzSchueler (approved):

PR #487, opened 2026-09-07
Make `GetMemoryFootprint` implementation consistent in `LRUCache`

ihtefyw:
The `LRUCache` implementation of `GetMemoryFootprint` and `GetDynamicMemoryFootprint` had multiple issues:
- They didn't include the map key allocations in the count
- `GetMemoryFootprint` returned the worst-case memory size instead of the current memory usage

This PR fixes those issues


go/common/lru_cache.go
```diff
@@ -193,18 +193,18 @@ func (c *LruCache[K, V]) dropLast() (dropped *entry[K, V]) {
 // If V is a pointer type, it needs to provide the size of a referenced value.
 // If the size is different for individual values, use GetDynamicMemoryFootprint instead.
 func (c *LruCache[K, V]) GetMemoryFootprint(referencedValueSize uintptr) *MemoryFootprint {
-	selfSize := unsafe.Sizeof(*c)
-	entrySize := unsafe.Sizeof(entry[K, V]{})
-	mf := NewMemoryFootprint(selfSize + uintptr(c.capacity)*(entrySize+referencedValueSize))
-	return mf
+	return c.GetDynamicMemoryFootprint(func(V) uintptr {
```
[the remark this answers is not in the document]
ihtefyw: This cache is not thread-safe and CachedHasher is not used anywhere

LuisPH3 (approved):

go/common/lru_cache_test.go
```diff
@@ -125,3 +126,38 @@ func TestCache_Entry_String(t *testing.T) {
...
+	cache := NewLruCache[int, int](3)
+	cache.Set(1, 11)
+	cache.Set(2, 22)
+
+	expectedSize := unsafe.Sizeof(*cache) +
+		3*(unsafe.Sizeof(int(0))+unsafe.Sizeof(&entry[int, int]{})) +
+		2*(unsafe.Sizeof(entry[int, int]{})+unsafe.Sizeof(int(0)))
+
+	require.Equal(expectedSize, cache.GetMemoryFootprint(unsafe.Sizeof(int(0))).Total())
+}
```
LorenzSchueler, line 219: This is a bit misleading. referencedValueSize is supposed to be the size of the allocation held by each value. Here the values are of type int, so it should be zero. Since you have `2*(...+unsafe.Sizeof(int(0)))` and pass `unsafe.Sizeof(int(0))` the test passes, but it is not computing the actual size. The easiest fix would be to use `&int` as the value type.

go/common/lru_cache_test.go
```diff
@@ -125,3 +126,38 @@ func TestCache_Entry_String(t *testing.T) {
...
+
+	expectedSize := unsafe.Sizeof(*cache) +
+		3*(unsafe.Sizeof(int(0))+unsafe.Sizeof(&entry[int, int]{}))
+	for key, size := range sizes {
+		cache.Set(key, key)
+		expectedSize += unsafe.Sizeof(entry[int, int]{}) + size
+	}
+
+	require.Equal(expectedSize, cache.GetDynamicMemoryFootprint(func(v int) uintptr {
+		return sizes[v]
```
LorenzSchueler: Same here. Here it would make sense to use a slice as the value type so that the values can actually have diffent sizes.

LorenzSchueler (changes requested):
The implementation is fine, the tests need some adjustment.

go/common/lru_cache_test.go
ihtefyw: that's true, I misunderstood the referencedValueSize field.
Fixed that

LorenzSchueler (approved):

PR #490, opened 2026-09-08
Declare the StagedBlock interface

ihtefyw:
Previously, applying a block was a final operation, hence all informations about it was lost after applying it to the `State`.
To enable block reverting, we need to retain informations about the block to be able to postpone the decision of accepting or reverting it. 
A simple solution to this comes from the live-archive structure of Carmen: the live state is volatile and stores only the latest trie values, while the archvie collects the history of all updates on the trie, which is updated once the live state udpate has terminated.  Therefore, we can consider the archive as the `committed` state, and the live state as a temporary store. 

The `StagedBlock` interface implements this concept. It represents a block that has been applied to the live state but still not serialized in the archive: it contains all informations required to committing it to the archive, or to revert it from the live state. Because it hasn't reached the archvie yet, no pruning is required.
The `Commit` and `Rollback` methods are mutually exclusive, and it's responsability of the user to respect this.


LorenzSchueler (commented):
It is hard to judge if this interface makes sense without seeing any of the implementation.
- Does a StagedBlock internally hold a reference to the DB, or how do Commit and Rollback work?
- Could Commit return a WaitHandle with a method Wait? That way it is impossible to call Wait before calling Commit.
- Would it make more sense do have StagedBlock being a pure data class, and have the Commit and Rollback methods on the DB and pass the StageBlock to them? This way the StagedBlock itself needs no DB reference. (Just an idea, not a suggestion).

ihtefyw:
@LorenzSchueler the idea is that the `StagedBlock` is what the user of Carmen sees after an update, and it contains everything it needs to commit and rolling back.

> Does a StagedBlock internally hold a reference to the DB, or how do Commit and Rollback work?
In the main implementation, yes is contains a reference to the `State`.
> Could Commit return a WaitHandle with a method Wait? That way it is impossible to call Wait before calling Commit.
    Would it make more sense do have StagedBlock being a pure data class, and have the Commit and Rollback methods on the DB and pass the StageBlock to them? This way the StagedBlock itself needs no DB reference. (Just an idea, not a suggestion).

ihtefyw:
@LorenzSchueler the new implementation does not have an explicit Wait() function

go/state/state.go
```diff
@@ -104,6 +111,75 @@ type State interface {
...
+// Rollback applies to the NEWEST, since every undo operation restores a value read
+// before its own block ran, and so reconstructs the intended state only once every
+// later block has already been rolled back.
+type StagedBlock interface {
+	// StateHash returns the root of the live state as of this block.
+	StateHash() common.Hash
+
+	// Commit promotes this block into the archive. It returns as soon as the write
+	// is under way, without waiting for it to complete; the returned handle is
+	// what waits for it. The handle is never nil when the error is.
```
LorenzSchueler:
```suggestion
	// is under way, without waiting for it to complete; the returned handle allows
	// waiting for it. The handle is never nil when the error is.
```

LorenzSchueler (approved):
The comments are a bit verbose, but I don't care too much about that. The interface itself is fine.

PR #493, opened 2026-09-08
Stop re-recording known errors on `Forest::flush`

ihtefyw:
`Forest.Flush` joined the already recorded errors (`CheckErrors`) with the errors of the current flush and appended the result back into the error list. Every flush after a single sticky error therefore nested the whole history once more, doubling the size of the recorded state each time: after 24 flushes the formatted error exceeds 200 MB, and a client that keeps flushing or logging `CheckErrors` runs out of memory instead of seeing a clean sticky error.

The early `CheckErrors` join dates back to the introduction of error tracking (#776), where `Flush` only returned it and never wrote back. The write-back was added later to make flush failures sticky (#1007) and unintentionally re-recorded the history along with them.

`Flush` now records only the errors produced by the flush itself and returns `CheckErrors`, so callers still receive all issues encountered so far, in the same order as before, and flush failures remain sticky.

Add a regression test that records one error through an injected stock and asserts that repeated flushes keep reporting it without growing the recorded error list.


go/database/mpt/forest_test.go
```diff
@@ -52,6 +53,73 @@ var forestConfigs = map[string]ForestConfig{
...
+	require.Len(t, forest.errors, 1)
+	size := len(forest.CheckErrors().Error())
+
+	// Subsequent flushes keep reporting the failure without recording it again.
+	for range 10 {
+		require.ErrorIs(t, forest.Flush(), injectedErr)
+		require.Len(t, forest.errors, 1)
+		require.Equal(t, size, len(forest.CheckErrors().Error()))
+	}
+}
```
LorenzSchueler: These tests differ by one line if im not mistaken. Can you use a single table test instead?

LuisPH3 (approved):
Thanks for comments

PR #497, opened 2026-09-10
Add the `IrreversibleBlock` implementation

ihtefyw:
An `irreversibleBlock` is a `StagedBlock` that doesn't support staging, and it's supposed to be used by `State` implementations that do not allow committing and rolling back nodes. 
Calling `Commit` is a no-op the first time, and an error the subsequent times to inform the user of interface misuse. Calling `Rollback` is always an error.


go/state/state.go
```diff
@@ -180,6 +182,42 @@ func (h *WaitHandle) Then(transform func(error) error) *WaitHandle {
...
 }
 
+// irreversibleBlock is a StagedBlock for a state that doesn't support staging.
+// Commit has nothing left to do but hand out a handle on whatever asynchronous work the state
+// started, reporting on the given channel, and Rollback reports that this state does not support it.
+//
+// It serves the state implementations that neither maintain an archive nor stage:
+// their staged sequence is always empty, so no ordering rule can be broken and
+// every block is final the moment it is applied.
+type irreversibleBlock struct {
```
LorenzSchueler, line 192: Isn't every state reversible using the undo list?

LorenzSchueler (approved):

go/state/state.go
ihtefyw: The `undoList` has been implemented only for the MPT, therefore S5.
All the other implementations do not support rolling back, but they go through the same interface

PR #498, opened 2026-09-10
Add the `stagedBlock` implementation to the Go state

ihtefyw:
This PR introduces the `stagedBlock`, the main mechanism for supporting block staging. 

After an `Apply`, a block is wrapped into a `stagedBlock` alongide other informations useful for committing/rollingback, such as a pointer to the `State` implementation, the update it was constructred from and the `undoList` containing the operations to revert it from the live state.
On `Commit`, the stored update is sent to the underlying archive, and on Rollback the `UndoList` is forwarded to the live state implementation.
The staged blocks are stored into the `GoState`, which enforce ordering on the commit/rolling back operations: only the newest block can be rolled back and only the oldest one can be committed.

A user receives a `StagedBlockHandle`, which is used by the `GoState` to uniquely identify a block. 
This implements the `StagedBlock` interface.

`GoState` currently hides this functionality by commiitting the block right away after applying it. 
The staging will be exposed in following PRs.


ihtefyw:
> stagedLock does not coordinate with the syncedState mutex used by normal state calls. Once a staged handle is exposed, Rollback can therefore execute RevertLastBlock concurrently with reads, Flush, or Apply. In particular, if another Apply has updated the LiveDB but has not yet appended its block to staged, this rollback still sees the previous block as newest and replays the wrong undo list over the new state. Handle decisions and the entire LiveDB apply/stage operation need to share the same synchronization domain (for example, an internal GoState lock acquired by every LiveDB access), rather than protecting only the queue.

This is gonna be handled when the staged block will be exposed

go/state/gostate/go_state.go
```diff
@@ -35,6 +35,18 @@ type GoState struct {
...
+	// oldest first, with everything either decision needs. A commit consumes the
+	// front, a rollback the back. A decision holds the lock from consuming
+	// its block to acting on it, so that the order in which blocks leave the
+	// queue is the order in which they reach the archive or are reverted.
+	// NOTE: this is unbounded, and may cause memory pressure.
+	staged     []*stagedBlock
+	stagedLock sync.Mutex
+	// stagedIds hands out the identity of each staged block, see stagedBlock.id.
+	// It is guarded by stagedLock.
+	stagedIds uint64
```
LorenzSchueler: This is not the number of staged blocks in `staged` but rather the highest id we have handed out so far.
```suggestion
	lastStagedId uint64
```

IMO `nextStagedId` would be even better. Then we have to increment after and not before assigning.

go/state/gostate/go_state.go
```diff
@@ -223,38 +237,237 @@ func (s *GoState) Apply(block uint64, update common.Update) (<-chan error, error
...
-	return archiveWriteDone, nil
+	handle.status = stagedRolledBack
+	return s.revertNewest()
+}
+
+// revertNewest takes the newest staged block back from the LiveDB. It must be
+// called while holding stagedLock, with at least one block staged.
+func (s *GoState) revertNewest() error {
+	last := len(s.staged) - 1
+	block := s.staged[last]
```
LorenzSchueler: Since `revertNewest` already return an error, it might be cleaner to check the length there instead of relying on the caller to do so.

go/state/gostate/go_state.go
```diff
@@ -223,38 +237,237 @@ func (s *GoState) Apply(block uint64, update common.Update) (<-chan error, error
...
+	s.stagedIds++
+	block.id = s.stagedIds
+	s.staged = append(s.staged, block)
+	return &stagedBlockHandle{state: s, id: block.id, hash: block.hash}
+}
+
+// commitStaged promotes the block the handle stands for into the archive. It
+// is rejected if the handle has been decided already or if its block is not the
+// oldest staged one.
+func (s *GoState) commitStaged(handle *stagedBlockHandle) (*state.WaitHandle, error) {
```
LorenzSchueler, line 314: This function also returns and error if there are errors in `s.stateError`. This contradicts the doc comment and makes it impossible to differentiate between a later block having failed or committing this block failed. Maybe don't check `getStateError` and leave this to `Check`?

go/state/gostate/go_state.go
```diff
@@ -223,38 +237,237 @@ func (s *GoState) Apply(block uint64, update common.Update) (<-chan error, error
...
 		}
-		if err := s.getStateError(); err != nil {
-			return nil, err
+	}()
+	return relay, nil
+}
+
+// stageBlock appends the block to the staged queue and returns the handle a
+// caller decides it with.
+func (s *GoState) stageBlock(block *stagedBlock) *stagedBlockHandle {
```
LorenzSchueler: Currently, there is nothing ensuring that `*stagedBlockHandle` implements `state.StagedBlock`.
```suggestion
func (s *GoState) stageBlock(block *stagedBlock) state.StagedBlock {
```


If we need the concrete type here we could also add this line somewhere
```
var _ state.StagedBlock = (*stagedBlockHandle)(nil)
```

go/state/gostate/go_state.go
```diff
@@ -223,38 +237,237 @@ func (s *GoState) Apply(block uint64, update common.Update) (<-chan error, error
...
+const (
+	stagedPending stagedStatus = iota
+	stagedCommitted
+	stagedRolledBack
+)
+
+// stagedBlockHandle is the StagedBlock a caller holds.
+// It's an opaque object identifying a staged block.
+type stagedBlockHandle struct {
+	state *GoState
```
LorenzSchueler, line 458: should the GoState be wrapped in `state.WrapIntoSyncedState(res)`?

LorenzSchueler (commented):
The ID based variant is definitely better than the variant before which relied on pointer equality.

go/state/gostate/go_state.go
ihtefyw: This is a pattern that happens a couple of times in the codebase, I think it's better to leave the error check to `Check()`

go/state/gostate/go_state.go
ihtefyw: No, the `syncedState` is gonna wrap both the `GoState` and the `StagedBlock`, it's gonna have its own implementation
Here they don't know about the synced state

go/state/gostate/go_state.go
```diff
@@ -223,43 +238,273 @@ func (s *GoState) Apply(block uint64, update common.Update) (<-chan error, error
...
+
+	if s.archive == nil {
+		return nil, nil // no archive, nothing to wait for
+	}
+	// Apply still reports the outcome of the archive write on a channel. The
+	// block was committed right here, so this is its only waiter.
+	relay := make(chan error, 1)
+	go func() {
+		defer close(relay)
+		if err := archiveWriteDone.Wait(); err != nil {
```
LuisPH3: how does this play along with the rest of the system? would a block which still rollback be used by the rpc as result of this notification
?

ihtefyw: An RPC would have to wait on the relay to know that the block was successfully added to the archive.
Importantly this is temporary: when the staged block will be exposed to the users, the carmen user will have to commit and wait on the block by himself

LuisPH3 (approved):

PR #501, opened 2026-09-16
Skip memory and flat impl & reduce test size in `stress_test.go`

ihtefyw:
The `stress_test`s were failing in CI because of memory exaustion. I reduced the test size by still keeping it representative of the test objective and skipped the memory and flat implementaiton, leaving only the production one running


LuisPH3 (approved):

LuisPH3:
pleas cherry pick to release branch, next release 2.2.x  can use this patch if needed.

PR #504, opened 2026-09-16
Expose `StagedBlock` through the `StateDB`

ihtefyw:
This completes the `StagedBlock` implementation by exposing the blocks outside of Carmen thorugh the `StateDB` interface. 
For reviewers: look first at the `StateDB` interface and `EndBlock` implementation, then proceed with the `stateDbStagedBlock` implementation, following by all the other tests.


go/carmen/block.go
```diff
@@ -85,10 +85,29 @@ func (c *headBlockContext) Commit() error {
 		}
 	}()
 
-	c.state.EndBlock(uint64(c.block))
+	// Ending the block only applies it to the live state. A block committed through
+	// this API is kept, so it is promoted into the archive as well.
+	staged, err := c.state.EndBlock(uint64(c.block))
+	if err == nil && staged == nil {
+		err = fmt.Errorf("state db ended block %d without returning a staged block", c.block)
```
LuisPH3: this code is very convoluted. can we simplify it? 
if you have to rewirite the the err value during your flow. this is an indication that something is wrong.

something like:
```
defer 	c.db.headStateCommitLock.Unlock()
if staged == nil && err == nil{
    return error blah blah 
}
``` 

use functions to create scopes, use defer for cleanups, avoid creating a statte machine with ifs

go/state/state_db.go
```diff
@@ -1318,43 +1319,56 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
...
-			go func() {
-				defer close(errRelay)
-				if syncError := archiveDone.Wait(); syncError != nil {
-					s.trackErrors(fmt.Errorf("failed to update archive for block %d: %w", block, syncError))
-					errRelay <- syncError
-				}
-			}()
-		}
-	}
-
```
LuisPH3: question, who does this now?

go/state/state_db.go
```diff
@@ -1318,43 +1319,56 @@ func (s *stateDB) EndBlock(block uint64) <-chan error {
...
+	block uint64
+}
+
+func (b *stateDbStagedBlock) Commit() (*WaitHandle, error) {
+	done, err := b.StagedBlock.Commit()
+	if err != nil {
+		return nil, err
+	}
+	// Only the outcome of the archive write is an issue of this StateDB.
+	return done.Then(func(err error) error {
```
LuisPH3: do we have guarantees that this terminates?

go/carmen/block.go
ihtefyw: Mhh the issue here is that EndBlock allows to return a nil StagedBlock (it's an interface) and a nil error, but that should never happen (either you get a block or an error). 
The best solution would have been to use a `Result` type, but unfortunately Go decided that `Result`s are a too advanced feature to bake them in the language. 

In your experience, how do you handle such a case in Go? I have a result type implemented in Carmen, but it's quite minimal and without the nice ergonomics you would get in Rust for example

go/state/state_db.go
ihtefyw: I am not sure I follow: this just decorates the `wait()` function, so it depends on if the wrapped `StagedBlock::wait()` terminates, which I don't think we can guarantee in any way as it's implementaiton-dependent.
We can only test the functions individually.

go/state/state_db.go
ihtefyw: The previous code was committing the block right ahead, while keeping the old interface returning a channel to wait for the archive synchronization
Now this is responsability of the user, i.e. Sonic: you call `EndBlock` and you get a `StagedBlock`. Calling `Commit + Wait` ensure you wait until the block is in the archive, which is the equivalent of waiting on the channel previously returned by `EndBlock`

go/state/state_db.go
ihtefyw: I am not sure I follow: this just decorates the wait() function, so it depends on if the wrapped StagedBlock::wait() terminates, which I don't think we can guarantee in any way as it's implementaiton-dependent.
We can only test the functions individually.

go/carmen/block.go
LuisPH3: https://github.com/0xsoniclabs/carmen/pull/506

go/state/state_db.go
LuisPH3: the logic is burried deep in the new infra. difficult to judge in the diff

go/state/state_db.go
LuisPH3: yes

go/carmen/block.go
```diff
@@ -75,19 +75,34 @@ func (c *headBlockContext) Commit() error {
...
-	c.db.headStateCommitLock.Lock()
-	headStateCommitLockReleased := false
-	defer func() {
-		// release the lock in case EndBlock panics.
-		if !headStateCommitLockReleased {
-			c.db.headStateCommitLock.Unlock()
+	// Ending the block only applies it to the live state. A block committed through
+	// this API is kept, so it is promoted into the archive as well. Both happen under
+	// exclusive (write) access to the head state, released before the head moves on.
+	err := func() error {
```
LuisPH3: nit: prefer not exported functions to lambdas, this is just complexity

LuisPH3 (approved):

PR #507, opened 2026-09-21
Simplify `GoState` error management

ihtefyw:
This PR improve and simplifies the error handling in the `GoState` module:
- Remove the `drainArchiveErrors` function: this could introduce a race between the archive background writer and anyone else calling it (e.g. `commitStaged`), where an error related to block X/Y could have been consumed by a function related to block Y/X
- Checks that the state is poisoned before attempting committing a block
- Return only the current error from `Get*` functions: they are already gated against poisoned state, it doesn't make sense to call `getStateError` again


go/state/gostate/go_state.go
```diff
@@ -69,32 +68,32 @@ func newGoState(live state.LiveDB, archive archive.Archive, cleanup []func()) st
...
 				} else {
-					// Otherwise, process the update.
+					// Otherwise, process the update. Its outcome is recorded here and
+					// reported to whoever waits for this write; it is this write's
+					// outcome alone, the health of the state is Check's business.
 					issue := res.archive.Add(update.block, *update.update, update.updateHints)
 					if issue != nil {
-						err <- issue
+						res.addStateError(issue)
 					}
```
LuisPH3: i  am a little confused by this followed by done channel pushing the same error. 
and in other location you see
```
	if s.archiveWriter != nil {
		// Wait until the flush was processed.
		s.addStateError(<-s.archiveWriterFlushDone)
	}
``` 
line 490

i like the returns changes, read better. but the introduction of the channel err is confusing. these are two separate asynchronous error handling methods which are somehow related.

LuisPH3 (approved):
thanks

